PR #7, opened 2018-08-09
Linode csi plugin added

sanjid133:
- [x] Complied with linode v4 api
- [x] Added CSI test


sanjid133:
cc: @tamalsaha

tamalsaha:
@displague , we have updated the FlexVolume to CSI driver. Also tests added using cs-test https://github.com/pharmer/csi-test/tree/linode

hack/deploy/releases/csi-linode-v0.0.1.yaml
```diff
@@ -109,27 +109,40 @@ spec:
...
+            - "--v=10"
           env:
             - name: CSI_ENDPOINT
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
             - name: LINODE_API_URL
               value: https://api.linode.com/
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
```
cliedeman: @displague This is very useful. I should be able to use this on the cloud controller side too to avoid using the hostname hack

PR #17, opened 2018-12-04
Add additional storage class with 'Retain' ReclaimPolicy

jfrederickson:
Kubernetes `PersistentVolume`s have an optional parameter that allows
you to define what happens to the volume after its deletion, and after
the deletion of a `PersistentVolumeClaim` that provisioned it. This
inherits from the `StorageClass` used to provision the volume.

The default setting results in the following behavior:

- A `PersistentVolume` is deleted when its `PersistentVolumeClaim` is
  deleted
- The underlying block storage volume is deleted when its
  `PersistentVolume` is deleted

This is normally what you would expect. However, in some cases this is
not desirable, for instance if the data contained within the volume is
particularly valuable and you don't want to risk inadvertently
deleting it.

By creating a second `StorageClass` using the `Retain` policy, we can
allow users to create volumes that are not deleted in either of the
above two cases. It's obviously possible for the cluster administrator
to configure an alternate `StorageClass` themselves, but I think
providing a `retain` option out of the box is useful (and at least one
other provider does so by default).

### DIFF
--- a/pkg/linode-bs/deploy/kubernetes/05-csi-storageclass.yaml
+++ b/pkg/linode-bs/deploy/kubernetes/05-csi-storageclass.yaml
@@ -6,3 +6,11 @@ metadata:
   annotations:
     storageclass.kubernetes.io/is-default-class: "true"
 provisioner: linodebs.csi.linode.com
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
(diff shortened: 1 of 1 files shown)


displague:
Is there precedence elsewhere for using an alias storage class rather than specifying the reclaimPolicy per use?

PR #22, opened 2019-02-01
fulfill volume requests using SP min -> req limit

displague:
The previous logic does not fulfill requests under the Linode size limit. 

This commit will use the Linode minimum size if the "required" size is too small to fulfill.

The "limit", if specified in a request, will be honored (as always) and is now preferred.

The spec for [CreateVolume](https://github.com/container-storage-interface/spec/blob/v1.0.0/spec.md#createvolume) reads:

 ```md
    // This field is OPTIONAL. This allows the CO to specify the capacity
    // requirement of the volume to be provisioned. If not specified, the
    // Plugin MAY choose an implementation-defined capacity range. If
    // specified it MUST always be honored, even when creating volumes
    // from a source; which MAY force some backends to internally extend
    // the volume after creating it.
    CapacityRange capacity_range = 2;

    ...

    // The capacity of the storage space in bytes. To specify an exact size,
    // `required_bytes` and `limit_bytes` SHALL be set to the same value. At
    // least one of the these fields MUST be specified.
    message CapacityRange {
      // Volume MUST be at least this big. This field is OPTIONAL.
      // A value of 0 is equal to an unspecified field value.
      // The value of this field MUST NOT be negative.
      int64 required_bytes = 1;

      // Volume MUST not be bigger than this. This field is OPTIONAL.
      // A value of 0 is equal to an unspecified field value.
      // The value of this field MUST NOT be negative.
      int64 limit_bytes = 2;
    }
 ```


displague:
Verified that `<10GB` and `>10GB` PVC requests were successfully fulfilled.

PR #39, opened 2020-02-06
Add logic to get instance by Linode ID

sibucan:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

**Fixes Issue**:
Fixes an issue where the CSI controller and CSI nodes pods would fail to be deployed when the kube node name was different from the Linode label. The fix was implemented by having the metadata service object to find a Linode instance by ID instead of by label.


app/linode/main.go
```diff
@@ -67,7 +68,22 @@ func handle() {
 	mounter := mountmanager.NewSafeMounter()
 	deviceUtils := mountmanager.NewDeviceUtils()
 
-	ms, err := metadataservice.NewMetadataService(cloudProvider, *node)
+	// Get information about the Linode this pod is executing in.
+	// Assume that the Linode ID file does not exist.
+	linodeInfo := *node
```
asauber: I think it would be cleaner to embed all of this logic inside `NewMetadataService`

pkg/linode-bs/deploy/kubernetes/06-ss-csi-linode-controller.yaml
```diff
@@ -44,7 +59,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: linode-csi-plugin
-          image: linode/linode-blockstorage-csi-driver:{{ .Values.image.tag }}
```
asauber: These manifests are a template for this script: https://github.com/linode/linode-blockstorage-csi-driver/blob/master/hack/release-yaml.sh

So, this value should be left as-is

pkg/linode-bs/deploy/kubernetes/08-cm-get-linode-id.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: ConfigMap
+metadata:
+  name: get-linode-id
+  namespace: kube-system
+  labels:
+    app: csi-linode
+data:
+  get-linode-id.sh: |-
+    #!/bin/bash -efu
+    id="$(kubectl get node/"${NODE_NAME}" -o jsonpath='{.metadata.annotations.lke\.linode\.com/linode-id}')"
```
asauber: Once the Linode CCM is fully initialized, `Node.spec.providerID` will also contain the Linode ID in the format: `linode://<LinodeID>`. This should look for that as well, as a fallback.

app/linode/main.go
sibucan: Agreed. I modified the code to match the original signature.

pkg/linode-bs/deploy/kubernetes/06-ss-csi-linode-controller.yaml
sibucan: Reverted the change 👍

pkg/linode-bs/deploy/kubernetes/08-cm-get-linode-id.yaml
sibucan: Realizing that the provider ID exists and how it is populated in the node object makes me believe that the `linode-id` annotation is probably at best redundant. I'll keep that in mind for future clean-up.

pkg/linode-bs/deploy/kubernetes/06-ss-csi-linode-controller.yaml
```diff
@@ -70,8 +85,18 @@ spec:
...
               mountPath: /var/lib/csi/sockets/pluginproxy/
       volumes:
         - name: socket-dir
           emptyDir: {}
+        - name: linode-info
+          emptyDir: {}
+        - name: get-linode-id
+          configMap:
+            name: get-linode-id
+            defaultMode: 493
```
asauber: I think these are always worth commenting

```suggestion
            # octal mode 755
            defaultMode: 493
```

pkg/linode-bs/deploy/kubernetes/08-cm-get-linode-id.yaml
```diff
@@ -0,0 +1,23 @@
...
+    if [[ ! -z "${id}" ]]; then
+      echo "Linode ID: "${id}"" 
+      echo -n ""${id}"" > /linode-info/linode-id
```
asauber: Do the double quotes in this string need to be escaped?

pkg/linode-bs/deploy/releases/linode-blockstorage-csi-driver-v0.1.4.yaml
```diff
@@ -0,0 +1,563 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: get-linode-id
+  namespace: kube-system
+  labels:
+    app: csi-linode
+data:
+  get-linode-id.sh: |-
+    #!/bin/bash -efu
```
asauber: You need to regenerate this with the new code

pkg/linode-bs/deploy/releases/linode-blockstorage-csi-driver.yaml
```diff
@@ -493,3 +543,21 @@ spec:
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: get-linode-id
+  namespace: kube-system
+  labels:
+    app: csi-linode
+data:
+  get-linode-id.sh: |-
+    #!/bin/bash -efu
```
asauber: This also needs to be regenerated using the new code

pkg/metadata/metadata.go
```diff
@@ -5,9 +5,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"strconv"
 
 	linodeclient "github.com/linode/linode-blockstorage-csi-driver/pkg/linode-client"
 	"github.com/linode/linodego"
+	"k8s.io/klog/glog"
```
asauber: If you are introducing klog, you should use `klog` and not the legacy `glog`

pkg/linode-bs/deploy/kubernetes/secret.yaml
```diff
@@ -1,8 +0,0 @@
-apiVersion: v1
-kind: Secret
-metadata:
-  name: linode
-  namespace: kube-system
-stringData:
-  token: "deadbeefab1e1ead__REPLACE_ME____deadbeefab1e1ead"
```
asauber: Please re-include this manifest. It is needed for external deployments.

pkg/linode-bs/deploy/kubernetes/08-cm-get-linode-id.yaml
```diff
@@ -0,0 +1,23 @@
...
+      exit 0
+    fi
+    id=kubectl get node/"${NODE_NAME}" -o jsonpath='{.spec.providerID}'
```
asauber: As we discussed, this script should probably _only_ use this field.

pkg/linode-bs/deploy/kubernetes/08-cm-get-linode-id.yaml
sibucan: No, since I am not trying to print the quotes themselves. I am just quoting the variable and then quoting the string that gets echo'd out, although just the quotes around echo seem like they should be enough. The double quotes don't hurt anyone in this case, however.

asauber: If you don't expect to print the inner double quotes, you should remove them. As is, this is three strings being concatenated together, as in the following example.

```
$ docker run -ti debian /bin/bash
root@a12bc3622668:/# a="hello"
root@a12bc3622668:/# echo -n ""${a}""
helloroot@a12bc3622668:/# echo -n "yes"${a}"ok"
yeshellookroot@a12bc3622668:/#
```

sibucan: Removed! 👍

pkg/linode-bs/deploy/kubernetes/06-ss-csi-linode-controller.yaml
```diff
@@ -16,6 +16,21 @@ spec:
         role: csi-linode
     spec:
       serviceAccount: csi-controller-sa
+      initContainers:
+        - name: init
+          image: bitnami/kubectl:1.17.2-debian-10-r11
```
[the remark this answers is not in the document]
asauber: Right now, 1.16 would be a better choice. See https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl

app/linode/main.go
```diff
@@ -67,7 +68,22 @@ func handle() {
...
+	isID := false
+
+	// Attempt to get the Linode ID information
+	data, err := ioutil.ReadFile("/linode-info/linode-id")
+	if err == nil {
```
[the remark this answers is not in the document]
sibucan: If the file does not exist or cannot be opened, the logic falls back to using the node name to find the linode. This is deliberate to support the use case where the CSI is deployed in a cluster without the Linode CCM, where this file is not expected to be present.

pkg/linode-bs/deploy/kubernetes/06-ss-csi-linode-controller.yaml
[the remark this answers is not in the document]
sibucan: Right, I can change this to be version 1.16.

hack/release-yaml.sh
```diff
@@ -1,13 +1,33 @@
...
+#!/bin/bash -eu
+set -o pipefail
+# Generate manifests for deployment on Kubernetes
 
-TAG=${1?$(git describe --abbrev=0)}
-RELEASES="pkg/linode-bs/deploy/releases/"
+# A tag name _must_ be supplied as the first argument
+TAG="${1}"
+if [[ -z "${TAG}" ]]; then
+    echo "Tag name to release must be supplied as the first argument"
```
asauber: To allow this error message to print, you will have to remove `-u` above

sibucan: Right! I can make that change.

README.md
```diff
@@ -79,10 +79,14 @@ Notably, this deployment will:
...
   ```
   
-* use a `reclaimPolocy` of `Released` [Learn More](https://kubernetes.io/docs/tasks/administer-cluster/change-pv-reclaim-policy/)
+* use a `reclaimPolicy` of `Released` [Learn More](https://kubernetes.io/docs/tasks/administer-cluster/change-pv-reclaim-policy/)
   
   Volumes created by this CSI driver will automatically be deleted when they are no longer needed.
 
+* assume that the [Linode CCM](https://github.com/linode/linode-cloud-controller-manager) is initialized and running [Learn More](https://kubernetes.io/docs/reference/command-line-tools-reference/cloud-controller-manager/)
+
+  If this is not the case, you must modify the init container script located in the `08-cm-get-linode-id.yaml` ConfigMap and delete [the line](https://github.com/linode/linode-blockstorage-csi-driver/blob/master/pkg/linode-bs/deploy/kubernetes/08-cm-get-linode-id.yaml#L18) that contains the `exit 1`.
```
asauber: "If this is not the case" -> "If you absolutely intend to run this on a cluster which will not run the Linode CCM"

pkg/linode-bs/deploy/kubernetes/08-cm-get-linode-id.yaml
```diff
@@ -0,0 +1,18 @@
...
+  get-linode-id.sh: |-
+    #!/bin/bash -efu
+    id="$(kubectl get node/"${NODE_NAME}" -o jsonpath='{.spec.providerID}')"
+    if [[ ! -z "${id}" ]]; then
+      echo "${id}"
+      echo -n "${id:9}" > /linode-info/linode-id
+      exit 0
+    fi
+    echo "Provider ID not found"
+    exit 1
```
asauber: Comment above this line with "# exit here so that we wait for the CCM to initialize the provider ID"

pkg/linode-bs/deploy/kubernetes/07-ds-csi-linode-node.yaml
```diff
@@ -39,7 +54,7 @@ spec:
               mountPath: /csi/
             - name: registration-dir
               mountPath: /registration/
-        - name: csi-linode-plugin 
+        - name: csi-linode-plugin
           image: linode/linode-blockstorage-csi-driver:{{ .Values.image.tag }}
           args :
             - "--endpoint=$(CSI_ENDPOINT)"
```
asauber: Let's take this opportunity to remove the `--v=10` here

pkg/metadata/metadata.go
```diff
@@ -27,10 +29,36 @@ type metadataServiceManager struct {
...
+		isID = true
+	}
+
+	// Linode instance
+	var linode *linodego.Instance
+	if isID {
+		// Search for Linode instance by ID
+		linode, err = getLinodeByID(linodeClient, linodeInfo)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to find Linode with ID %s: %v", linodeInfo, err)
```
asauber: You can simply return `nil, err` here, because you have already created this Error struct.

pkg/metadata/metadata.go
```diff
@@ -27,10 +29,36 @@ type metadataServiceManager struct {
...
+		}
+	} else {
+		// Search for Linode instance by label
+		linode, err = getLinodeByLabel(linodeClient, linodeInfo)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to find Linode with label %s: %v", linodeInfo, err)
```
asauber: Same comment as line 54

pkg/linode-bs/deploy/releases/linode-blockstorage-csi-driver.yaml
```diff
@@ -491,5 +524,23 @@ spec:
...
+    id="$(kubectl get node/"${NODE_NAME}" -o jsonpath='{.spec.providerID}')"
+    if [[ ! -z "${id}" ]]; then
+      echo "${id}"
+      echo -n "${id:9}" > /linode-info/linode-id
+      exit 0
+    fi
+    echo "Provider ID not found"
+    exit 1
```
asauber: With the above change you will need to regenerate this manifest.

PR #73, opened 2021-06-23
fix(k8s): apply standard tolerations for system daemonsets

themightychris:
It is standard practice for provider- and kubernetes-provided critical system daemonsets to come with wildcard tolerations.

Without these, any use of taints on nodes to shape workloads leads to nodes that fail to attach storage.

Even worse, because LKE doesn't support pre-assigning taints to nodepools, you get race conditions because taints have to be applied manually as nodes are added/recycled and come online. So you end up with storage working on a node initially and then randomly failing sometime later after the storage controller crashes and can't restart on any nodes with taints applied, and then it never working on some nodes if you managed to apply taints before the controller got scheduled.


g1bl3t (approved):
Looks good, thanks Chris!

sibucan (approved):

PR #77, opened 2021-12-02
Update e2e tests

sibucan:
### General:

### Pull Request Guidelines:

1. [X] Does your submission pass tests?
1. [ ] Have you added tests? 
1. [X] Are your commits atomic, addressing one change per commit?
1. [X] Are you following the conventions of the language? 
1. [X] Have you explained your rationale for why this feature is needed? 

---

Previously, the e2e tests would:
1. Not fail immediately if ran directly (without the makefile) and the `LINODE_API_TOKEN` was not set as an environment variable.
2. Use `time.Sleep()` to "wait" for BS Volumes to be detached/deleted
3. Use hardcoded values for timeouts and retry intervals.
4. Pass the StorageClass name as a input to the framework, instead of being a per-test parameter

The changes in here address those issues:
1. The tests themselves now fail quickly if the API token is not provided, or if the Kubernetes version is not provided and a cluster needs to be created.
2. The tests now "poll" to check if the volumes have detached and deleted.
3. The timeouts and retry intervals can be modified using the `--timeout` flag and the `--retry-interval` flag, respectively.
4. Has the `linode-block-storage` StorageClass name not be a framework parameter.

Additionally, I also took the opportunity to update linodego and added extra flags to configure its runtime behavior as well.


Charliekenney23 (approved):
Looks great! Runs much more quickly.

Issue #63, opened 2021-01-02
Support Volume Encryption


kekoav:
## Feature Request

Support volume encryption for LKE clusters via the CSI driver.

## Reasoning

Data encryption is critical for ensuring the security of data at rest for any system. Volume encryption is a simple method to ensure all data is protected if physical security of the hardware is ever compromised. Encryption is now a standard feature of modern clouds, and it is required for PCI-DSS compliance and other security standards.

## Other Solutions

Other cloud providers provide volume encryption as an option within the StorageClass.

* [AWS](https://kubernetes.io/docs/concepts/storage/storage-classes/#aws-ebs) provides StorageClass option
* [IBM Cloud](https://cloud.ibm.com/docs/containers?topic=containers-block_storage) provides StorageClass option
* [DigitalOcean](https://www.digitalocean.com/docs/volumes/#features) encrypts with LUKS by default
* [Google Cloud](https://cloud.google.com/kubernetes-engine/docs/how-to/using-cmek) encrypts by default and manages keys, but also provides the option for customers to manage keys

## Proposed Solution

I propose a solution which enables encryption simply via the StorageClass definition using Linode-managed keys for simplicity for the users, but also allows for unencrypted volumes for use cases where users do not need encryption.

### Example StorageClass With `encrypted` Attribute
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    lke.linode.com/caplke-version: v1.x
  name: linode-block-storage-encrypted
allowVolumeExpansion: true
encrypted: true
provisioner: linodebs.csi.linode.com
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

## Implementation Ideas

The simplest way seems to be with LUKS support similar to DigitalOcean. Key management is going to be important, and how they are managed should be transparent to users. The security of these keys requires some type of KMS.

xxated:
I would find that extremely useful as well!

PR #85, opened 2022-10-11
Feature: Adds optional support for LUKS encrypted mounts with an ext4 filesystem with a user supplied key

kitknox:
New functionality adding the ability to encrypt a PVC with a user owned secret provides an additional security layer that gives control of the data to the cluster owner instead of the platform provider.  Secret management is out of the scope of the CSI driver itself and currently expects a native Kubernetes secret as the source for a given volume.

This new functionality is skipped if the optional luks-encrypted: "true" is not present for the StorageClass.  You can have a mix of StorageClasses with and without LUKS enabled.

Tested on various LKE and non-LKE clusters with Linode CCM and CSI driver installed.


displague:
Would this resolve #63?

kitknox:
That specifically proposes a case where Linode holds/manages the encryption keys.  This implementation has the customer holding the keys which is both a pro and a con for different customers.  We can't open a customers data when we don't have the key, but a customer now has to own keeping that key secure if they manage it.

frenchtoasters:
Thanks for adding this feature @kitknox would you also be able to add updated tests to this PR for this new feature?

mfechtner:
Hi, appreciate your efforts, any plans to make this usable soon?
@frenchtoasters @kitknox

pkg/linode-bs/luks.go
```diff
@@ -0,0 +1,405 @@
+/*
+Copyright 2019 linkyard ag
+Copyright cloudscale.ch
+Copyright 2022 Akamai Technologies
```
luthermonson, line 4: since unfortunately they didnt export any of this stuff and you had to lift the file can you least link to the source https://github.com/cloudscale-ch/csi-cloudscale/blob/master/driver/luks_util.go and give them credit in a comment?

pkg/linode-bs/luks.go
```diff
@@ -0,0 +1,405 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package linodebs
+
+import (
```
luthermonson, line 21: this whole file is a mess and i know you got it that way but let's run goimports across is at clean it up a bit

pkg/linode-bs/nodeserver.go
```diff
@@ -30,6 +30,8 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/utils/mount"
+	"os/exec"
```
luthermonson, line 33: goimports

pkg/linode-bs/nodeserver.go
```diff
@@ -197,6 +239,8 @@ func (ns *LinodeNodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeSt
 
 	glog.V(4).Infof("Successfully found attached Linode Volume %q at device path %s.", deviceName, devicePath)
 
+	luksContext := getLuksContext(req.Secrets, req.VolumeContext, VolumeLifecycleNodeStageVolume)
```
luthermonson, line 242: can this move closer to ln 283 where it starts getting used?

pkg/linode-bs/nodeserver.go
```diff
@@ -236,11 +280,37 @@ func (ns *LinodeNodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeSt
 		return nil, status.Error(codes.Unimplemented, "Block volume support is not yet implemented")
 	}
 
-	err = ns.Mounter.FormatAndMount(devicePath, stagingTargetPath, fstype, options)
-	if err != nil {
+	glog.V(4).Info("luksContext encrypted ", luksContext.EncryptionEnabled)
+	formatted, err := blkidValid(devicePath)
+	glog.V(4).Info("luks volume formatted  ", formatted)
+	if (!formatted && luksContext.EncryptionEnabled) {
```
luthermonson, line 286: need to clean up the conditionals a bit, no need for two encryption enabled checks etc.

luthermonson: thinking about this a bit more, i'd check if encryption is enalbed before even calling `getLuksContext` or `blkidValid`... none of it should be called if we're not even using encryption

pkg/linode-bs/nodeserver.go
```diff
@@ -44,6 +46,7 @@ type LinodeNodeServer struct {
 	MetadataService metadata.MetadataService
 	// TODO: Only lock mutually exclusive calls and make locking more fine grained
 	mux sync.Mutex
+	luksContext	LuksContext
```
luthermonson, line 49: i dont think this is ever used?

luthermonson:
moved to https://github.com/linode/linode-blockstorage-csi-driver/pull/122

PR #89, opened 2023-01-17
Update e2e_test go and ginkgo

mrogers950:
Update the e2e_test module go to 1.17 and bump ginkgo/gomega to align with https://github.com/linode/linode-cloud-controller-manager/pull/110.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


mrogers950:
I've confirmed that the tests pass with this change.

e2e/test/scripts/create_cluster.sh
```diff
@@ -7,7 +7,8 @@ set -x
 
 export LINODE_API_TOKEN="$1"
 export CLUSTER_NAME="$2"
-export K8S_VERSION="$3"
+export REGION="$3"
+export K8S_VERSION="$4"
```
git-it127, line 11: In order to not break any possible existing automation I think we should try to keep the params in the same order. So we can try this instead.

```suggestion
export K8S_VERSION="$3"
export REGION="$4"
```

e2e/test/scripts/create_cluster.sh
```diff
@@ -22,7 +23,7 @@ variable "server_type_master" {
   default = "g6-standard-2"
 }
 variable "region" {
-  default = "eu-west"
+  default = "$REGION"
```
git-it127, line 26: Again to not break any existing automation maybe we can look to check if `REGION` is not set. If not set then keep this default. If it is set then override it with what was passed in.

git-it127 (changes requested):
Overall looks like some good updates. Have a few small suggestions here.

e2e/test/scripts/create_cluster.sh
mrogers950: Done, thanks for the suggestion.

mrogers950:
@git-it127 I've updated with your suggestions.

git-it127 (approved):

PR #118, opened 2023-10-03
v0.5.3 Release

luthermonson:
* pinning versions for v0.5.3 release
* updating docs


kishorchannal (approved):
LGTM

RELEASES.md
git-it127, line 1: Rather than deleting this file. I think it would make more sense to make a RELEASING doc with updated instructions. That would include a lot of the steps here like updating the manifests and helm charts.

git-it127 (changes requested):

RELEASES.md
luthermonson: i hadn't gone through the process of the tagging to confirm everything works, was planning that a releasing doc to be a followup PR later today after everything was out but that was only something in my head and not expressed and not really expressed any where in this PR. could we do a place holder it with the basics with a todo for a followup after the release?

git-it127: Yeah that sounds good to me. You could do the full thing after the release if you need to test out the process first. Just as long as the doc is updated.

git-it127 (approved):
LGTM

PR #127, opened 2023-10-22
adding windows build tags to hide unix.Statfs calls

luthermonson:
Getting `NodeGetVolumeStats` implemented added a dependency to `golang.org/x/sys/unix` which will never work on windows nodes. This driver does NOT support Windows nodes but this does make running the unit tests on a windows desktop work and as an open source project we should acknowledge development may happen on any platform.

## Running `TestDriverSuite` on a win11 desktop from Goland
**Before changes:** 
```
# github.com/linode/linode-blockstorage-csi-driver/pkg/linode-bs [github.com/linode/linode-blockstorage-csi-driver/pkg/linode-bs.test]
.\nodeserver.go:392:18: undefined: unix.Statfs_t
.\nodeserver.go:394:14: undefined: unix.Statfs
.\nodeserver.go:396:26: undefined: unix.ENOENT
```

**After changes:**  
```
=== RUN   TestDriverSuite
--- PASS: TestDriverSuite (0.00s)
PASS
```

### DIFF
--- a/pkg/linode-bs/nodeserver_all.go
+++ b/pkg/linode-bs/nodeserver_all.go
@@ -1,0 +1,46 @@
+//go:build !windows
+
+package linodebs
+
+import (
+	"context"
+	"errors"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"golang.org/x/sys/unix"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func nodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
+	if req.VolumeId == "" || req.VolumePath == "" {
+		return nil, status.Error(codes.InvalidArgument, "volume ID or path empty")
+	}
+
+	var statfs unix.Statfs_t
+	// See http://man7.org/linux/man-pages/man2/statfs.2.html for details.
+	err := unix.Statfs(req.VolumePath, &statfs)
+	if err != nil {
+		if errors.Is(err, unix.ENOENT) {
+			return nil, status.Errorf(codes.NotFound, "volume path not found: %v", err.Error())
+		}
+		return nil, status.Errorf(codes.Internal, "failed to get stats: %v", err.Error())
+	}
+
+	return &csi.NodeGetVolumeStatsResponse{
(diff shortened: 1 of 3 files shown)


go.mod
schinmai-akamai, line 1: What’s causing the go.mod changes?

luthermonson: We commented out the csi tests on Friday since they weren't doing anything during ci, will readd when we get e2e working again

okokes-akamai: Can we then split the PR to not include this change to keep this cleaner?

pkg/linode-bs/nodeserver_all.go
```diff
@@ -0,0 +1,47 @@
+//go:build !windows
+// +build !windows
```
okokes-akamai, line 2: Do we want to support legacy build tags? `go fix` will remove these anyway and I don't think we support a toolchain version that doesn't know about `go:build`

pkg/linode-bs/nodeserver_windows.go
```diff
@@ -0,0 +1,14 @@
...
+	"context"
+	"fmt"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func nodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
+	return nil, status.Error(codes.Unimplemented, fmt.Sprintf("NodeGetVolumeStats is not yet implemented"))
```
okokes-akamai, line 13: Can we perhaps add something like "not yet implemented *on Windows*", so that the error is a bit clearer?

go.mod
luthermonson: sure np

pkg/linode-bs/nodeserver_windows.go
luthermonson: great idea!

pkg/linode-bs/nodeserver_all.go
luthermonson: default for goimports fixes this i believe... pretty sure i only put in the +build and it added it for me. let me investigate but I agree with you, probably no reason for both at this point

schinmai-akamai (approved):

okokes-akamai (approved):

PR #133, opened 2023-11-08
Allow passing tags to volumes using the storage class

### LINKED ISSUE #132: Add Tag Support for Volumes
When creating a volume we should be able to pass in meta data to the PVC to push tags into the linode api for the created volume.

schinmai-akamai:
### General:

Adds a new feature to pass tags to linode volumes created by the CSI driver. It is currently not possible to set an annotation on the PVC and get the csi-driver to read the annotation due to https://github.com/kubernetes-csi/external-provisioner/issues/86

Choosing to take the approach of passing tags via StorageClass parameters instead.

This closes #132 

### Pull Request Guidelines:

1. [ ] Have you added tests? 


pkg/linode-bs/luks.go
```diff
@@ -257,7 +261,7 @@ func luksClose(volume string) error {
 func luksOpen(volume string, keyFile string, ctx LuksContext) error {
 	// check if the luks volume is already open
 	if _, err := os.Stat("/dev/mapper/" + ctx.VolumeName); !os.IsNotExist(err) {
-		glog.V(4).Info("luks volume is already open %s", volume)
+		glog.V(4).Infof("luks volume is already open %s", volume)
```
schinmai-akamai, line 264: linter was complaining, figured i'd fix it too

luthermonson:
add to the howto in the readme how to do this without the helm chart?

schinmai-akamai:
Tested by installing the helm chart with allowTags set to `awjv0d0`
 
```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  creationTimestamp: "2023-11-09T01:47:47Z"
  name: linode-block-storage-retain
  resourceVersion: "4756"
  uid: a68363b3-68d6-49d8-b748-0d08dff099ba
parameters:
  linodebs.csi.linode.com/volumeTags: awjv0d0
provisioner: linodebs.csi.linode.com
reclaimPolicy: Retain
volumeBindingMode: Immediate
```

I've got a teleport pv created
```
❯ k get pv pvc-322d8f57747c4922  -n kube-system

NAME                   CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                             STORAGECLASS                  REASON   AGE
pvc-322d8f57747c4922   10Gi       RWO            Retain           Bound    kube-system/teleport-in-cluster   linode-block-storage-retain            7m13s
```

Querying the Linode API for volumes - you can see that the volume has a tag on it
```
{
    "id": 1959917,
    "status": "active",
    "label": "pvc322d8f57747c4922",
    "created": "2023-11-09T01:50:10",
    "updated": "2023-11-09T01:50:35",
    "filesystem_path": "/dev/disk/by-id/scsi-0Linode_Volume_pvc322d8f57747c4922",
    "size": 10,
    "linode_id": 51795491,
    "linode_label": "rags-worker-0",
    "region": "us-east",
    "tags": [
      "awjv0d0"
    ],
    "hardware_type": "nvme"
  },
```

schinmai-akamai:
> add to the howto in the readme how to do this without the helm chart?

Yes - added

pkg/linode-bs/controllerserver.go
```diff
@@ -102,6 +102,8 @@ func (linodeCS *LinodeControllerServer) CreateVolume(ctx context.Context, req *c
 		volumeContext[LuksKeySizeAttribute] = req.Parameters[LuksKeySizeAttribute]
 	}
 
+	tags := req.Parameters[VolumeTags]
```
srust, line 105: It is weird that VolumeTags comes from the LUKS constants

helm-chart/csi-driver/templates/linode-block-storage.yaml
```diff
@@ -7,5 +7,9 @@ metadata:
   annotations:
     storageclass.kubernetes.io/is-default-class: "true"
 {{- end }}
+{{- if .Values.addTags }}
+parameters:
+  linodebs.csi.linode.com/volumeTags: "{{ .Values.addTags }}"
```
srust, line 12: can this simply be "{{ .Values.volumeTags }}" instead of "addTags" , to match the parameter name?

pkg/linode-bs/controllerserver.go
srust, line 105: Also I'm sure this is tested, but can `req.Parameters[VolumeTags]` return nil for the tags?

I don't think we check and `strings.split` is called directly later on on this tags variable that cannot be nil

srust (changes requested):

pkg/linode-bs/controllerserver.go
schinmai-akamai: req.Paramaters here is the params on the storage class

schinmai-akamai: ah the const is defined in luks to keep things in one place - can move it here if it makes more sense

schinmai-akamai: https://go.dev/ref/spec#Index_expressions

```
For a of [map type](https://go.dev/ref/spec#Map_types) M:

x's type must be [assignable](https://go.dev/ref/spec#Assignability) to the key type of M
if the map contains an entry with key x, a[x] is the map element with key x and the type of a[x] is the element type of M
if the map is nil or does not contain such an entry, a[x] is the [zero value](https://go.dev/ref/spec#The_zero_value) for the element type of M
```

I was relying on #3 in the above to guarantee a non-nil string

srust: I'd move the few non LUKS constants to here and keep the LUKS stuff in LUKS. The drivername and the volume tags and phases I think should be in this file.

schinmai-akamai: moved it over

luthermonson: i did this math when we added the luks stuff, map[string]string will return an empty string, not nil: https://go.dev/play/p/myxllIPEiqe

the other stuff like strings.Split will be fine with an empty string input.

helm-chart/csi-driver/values.yaml
```diff
@@ -18,7 +18,11 @@ namespace: kube-system
 # "linode-block-storage" or left as an empty string
 defaultStorageClass: linode-block-storage-retain
 
-# Images - Default
+# set these value to a comma seperated string if you'd like to add tags to the created volumes
+# volumeTags:
```
luthermonson, line 22: nit: might be a nicer interface for a helm chart to do a list and `{{ join "," .Values.volumeTags }}` ?

```
volumeTags:
 - tag1
 - tag2
```

luthermonson (approved):

helm-chart/csi-driver/values.yaml
srust: That would be nice..

srust (approved):

helm-chart/csi-driver/values.yaml
schinmai-akamai: done - updated

PR #138, opened 2023-12-06
Lint e2e, install ginkgo, downgrade apimachinery

okokes-akamai:
Several improvements to our e2e tooling:

1. Lint e2e tests - to uncover problems with these tests not even compiling (not talking about running them) - this _will_ fail CI
2. Install ginkgo if not installed already - this will allow us to run `make test` for end to end tests, something we've not been able to
3. Downgrade apimachinery, which is too new for our e2e tests to work - without this our tests won't compile


mabojars:
Do we have effort planned for addressing the issues uncovered by lint?

luthermonson (approved):

okokes-akamai:
@mabojars that's happening now

PR #140, opened 2023-12-06
Fixing E2E Tests

luthermonson:
Core changes to get these running again and we should pass linting again. To get these running you can boot an LKE cluster and edit the image in the stateful/daemon set for the csi-driver to be an image you build/push to your own image repo. More updates coming to automate that a bit more.

* changed default interval to 1s to make tests run faster, still more to do but this helped cut off 2-3min
* refactored pods/pvcs to have an random int in their name, made them easier to track in kube while watching tests and hopefully we can run them all in parallel some day soon
* found dupe tests running two cleanups
* hardened the funcs around delete/detach to return errors if 404s/notfound/etc
* refactored raw block tests to be a part of the Pod + PVC test
* refactored mkfs.ext4 func to be like the other ExecInPods
* updated all tests to run in ubuntu for raw block to have mkfs.ext4 and consolidated into one GetPodObject func

### DIFF
--- a/e2e/test/csi_driver_test.go
+++ b/e2e/test/csi_driver_test.go
@@ -3,8 +3,8 @@ package test
 import (
 	"e2e_test/test/framework"
 	"fmt"
+	"math/rand"
 	"strconv"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -19,11 +19,11 @@ var _ = Describe("Linode CSI Driver", func() {
 		Context("Using VolumeClaimTemplates and a non-root container", func() {
 			var (
 				f            *framework.Invocation
-				storageClass = "linode-block-storage"
 				sts          *appsv1.StatefulSet
 				pod          *core.Pod
-				// pvc          *core.PersistentVolumeClaim
-				// file         = "/data/file.txt"
+				pvc          *core.PersistentVolumeClaim
+				file         = "/data/file.txt"
+				storageClass = "linode-block-storage"
 			)
 
 			BeforeEach(func() {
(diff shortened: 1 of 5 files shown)


e2e/test/csi_driver_test.go
```diff
@@ -61,33 +61,71 @@ var _ = Describe("Linode CSI Driver", func() {
...
+				f    *framework.Invocation
+				size string
+				// storageClass = "linode-block-storage"
+				sts          *appsv1.StatefulSet
+				pod          *core.Pod
+				pvc          *core.PersistentVolumeClaim
+				file         = "/data/file.txt"
+				storageClass = "linode-block-storage"
+
+				writeFile = func(filename string) {
```
okokes-akamai, line 74: It's a bit odd to duplicate this here (same for `readFile`).

e2e/test/csi_driver_test.go
```diff
@@ -61,33 +61,71 @@ var _ = Describe("Linode CSI Driver", func() {
 
 	Describe("Test", func() {
 		Context("Simple", func() {
+			var (
+				f    *framework.Invocation
```
okokes-akamai, line 65: Can/should we share some (all?) of these atop of this group of tests?

e2e/test/csi_driver_test.go
```diff
@@ -61,33 +61,71 @@ var _ = Describe("Linode CSI Driver", func() {
...
-					err = f.DeletePod(pod.Name)
+					err := f.DeletePod(pod.ObjectMeta)
 					Expect(err).NotTo(HaveOccurred())
 
 					By("Waiting for the Volume to be Detached")
-					waitForOperation()
+					volumeID, err := f.GetVolumeID(pvc)
+
+					Eventually(func() bool {
+						isAttached, err := f.IsVolumeDetached(volumeID)
```
okokes-akamai, line 108: `isAttached, _ = f.IsVolumeDetached`?

e2e/test/csi_driver_test.go
```diff
@@ -61,33 +61,71 @@ var _ = Describe("Linode CSI Driver", func() {
...
 
 					By("Deleting the PVC")
 					err = f.DeletePersistentVolumeClaim(pvc.ObjectMeta)
 					Expect(err).NotTo(HaveOccurred())
 
 					By("Waiting for the Volume to be Deleted")
-					waitForOperation()
+					Expect(err).NotTo(HaveOccurred())
+
+					Eventually(func() bool {
```
okokes-akamai, line 122: This swallows the error, makes debugging close to impossible.

e2e/test/framework/pod.go
```diff
@@ -108,6 +109,16 @@ func (f *Invocation) WaitForReady(meta metav1.ObjectMeta) error {
 	})
 }
 
+func (f *Invocation) WaitForDelete(meta metav1.ObjectMeta) error {
+	return wait.PollImmediate(f.RetryInterval, f.Timeout, func() (bool, error) {
+		_, err := f.kubeClient.CoreV1().Pods(meta.Namespace).Get(meta.Name, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			return false, nil
```
okokes-akamai, line 116: `true`?

e2e/test/framework/pod.go
```diff
@@ -108,6 +109,16 @@ func (f *Invocation) WaitForReady(meta metav1.ObjectMeta) error {
...
+		}
+		return false, nil
```
okokes-akamai, line 118: `, err`?

e2e/test/csi_driver_test.go
luthermonson: ya totally agree, thinking about coalescing too! ill do this after i get them running

e2e/test/csi_driver_test.go
luthermonson: someone decided these needed to be scoped to the Describe... i dont know if i can properly answer this question yet but I agree with you and ill give it a try and see where it comes out

e2e/test/csi_driver_test.go
luthermonson: oh ya that's cleaner lemme do that

e2e/test/csi_driver_test.go
luthermonson: all of them were written this way so i followed suite, can we agree to make an issue to rewrite these better and not include this as a part of getting these things semi functional again?

e2e/test/framework/pod.go
luthermonson: derp

e2e/test/framework/pod.go
luthermonson: ya that would work well if im catching the is not found, good idea

e2e/test/csi_driver_test.go
luthermonson: ah but to your other point, we could dump these errors for debugging purposes?

okokes-akamai: I'm not saying to discard the error, I was pointing to the fact that the variable is named `isAttached` but the method presumably returns the opposite.

e2e/test/csi_driver_test.go
okokes-akamai: Why can't we just write them properly in the first place? I only commented this one, because it looks like it's a new one, I'm not suggesting we redo all of them.

(The ticket to refactor these is already in place and I'm working on it - I got stopped by these tests being broken.)

e2e/test/framework/pod.go
```diff
@@ -42,6 +38,26 @@ func GetPodObject(name, namespace, pvc string) *core.Pod {
 			},
 		},
 	}
+
+	if volumeType == core.PersistentVolumeFilesystem {
```
okokes-akamai, line 42: Exhaustive switch instead of two `if`s?

e2e/test/framework/pod.go
```diff
@@ -88,7 +104,11 @@ func (f *Invocation) CreatePod(pod *core.Pod) error {
 }
 
 func (f *Invocation) DeletePod(meta metav1.ObjectMeta) error {
-	return f.kubeClient.CoreV1().Pods(meta.Namespace).Delete(meta.Name, deleteInForeground())
+	err := f.kubeClient.CoreV1().Pods(meta.Namespace).Delete(meta.Name, deleteInForeground())
+	if err != nil && apierrors.IsNotFound(err) {
```
okokes-akamai, line 108: Does this version of `IsNotFound` not accept a nil error? If so, why are you using this function in `WaitForDelete` without the `if err != nil` gate?

avestuk (approved):

e2e/test/e2e_suite_test.go
```diff
@@ -36,7 +36,7 @@ func init() {
 	flag.BoolVar(&useExisting, "use-existing", useExisting, "Use an existing kubernetes cluster")
 	flag.StringVar(&framework.KubeConfigFile, "kubeconfig", filepath.Join(homedir.HomeDir(), ".kube/config"), "To use existing cluster provide kubeconfig file")
 	flag.DurationVar(&framework.Timeout, "timeout", 5*time.Minute, "Timeout for a test to complete successfully")
-	flag.DurationVar(&framework.RetryInterval, "retry-interval", 5*time.Second, "Amount of time to wait between requests")
+	flag.DurationVar(&framework.RetryInterval, "retry-interval", 1*time.Second, "Amount of time to wait between requests")
```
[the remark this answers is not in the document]
luthermonson: im going to back this change out, at the beginning when the tests were all broke and retrying to 404s and stuff because the cleanups were being double ran this improved test speeds by 2-3min. today with more or less working tests I tested 1s vs 5s retries and it was only a 30s faster and developers and this is a meaningless change at this point.

e2e/test/framework/pod.go
luthermonson: so the bigger problem with WaitForDelete is it takes an ObjectMeta and assumes it's a pod, its supposed to be deleted and I'll remove some unused code in separate commit.

RE the actual question, it is fine to not have the nil check and ill rewrite DeletePod accordingly.

e2e/test/framework/pod.go
luthermonson: the thing I know i wanted is passing in a required non-pointer volume type to pod creation and this accomplished that so we can write a `Context()` to set volumeType like [this](https://github.com/linode/linode-blockstorage-csi-driver/pull/140/files#diff-7bc5b2d13fd2086f3a76c2fc154cfdefc57ed7f8e3964031c5c8936fa76c2876R273). can we leave this for the moment? i believe a follow up task to add to the block tests to have some setup that makes them behave like the filesystem is coming and this will be refactored in the process. if i refactored this today it'll just be a switch with two cases and if you want that let me know.

e2e/test/csi_driver_test.go
luthermonson: i dont think this is valid feedback any more after rewriting stuff, im going to resolve and you can open a new one against anything in the current pr

lgarber-akamai:
@luthermonson Given the description mentions running this against an LKE cluster, do you plan on removing the [dependency on terraform-linode-k8s](https://github.com/linode/linode-blockstorage-csi-driver/blob/2d8467e220f4cc039961f174c5b464d458c15061/e2e/test/e2e_suite_test.go#L86-L91)?

luthermonson:
> @luthermonson Given the description mentions running this against an LKE cluster, do you plan on removing the [dependency on terraform-linode-k8s](https://github.com/linode/linode-blockstorage-csi-driver/blob/2d8467e220f4cc039961f174c5b464d458c15061/e2e/test/e2e_suite_test.go#L86-L91)?

ya when we move to booting an LKE cluster in the next pr this will probably get pulled

e2e/test/framework/pod.go
okokes-akamai: I just want this to err (?) if it's not one of these two types as that seems unsupported (?)

luthermonson: that's a totally fair ask. i rewrote as a switch and added error returns for both the pod and pvc funcs and made them look/behave similarly so they feel more cohesive. let me know if you approve these changes: https://github.com/linode/linode-blockstorage-csi-driver/pull/140/commits/c63960a2a0a1b04a6bcc18c8b54e1580f0928a8e

okokes-akamai (approved):

avestuk (approved):

PR #141, opened 2023-12-06
forgot InitFlags

luthermonson:
### General:

* forgot to commit the init flags in main

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,6 +44,7 @@ func init() {
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


luthermonson:
linter will be broken until #140 is done and this needs to be force merged

lgarber-akamai (approved):
Tested in an RKEv2 cluster and this seems to resolve the missing `-v` flag error 🙂

schinmai-akamai (approved):

PR #142, opened 2023-12-07
ref: Respect idempotency in CreateVolume(...)

lgarber-akamai:
### General:

This change updates the `LinodeControllerServer.CreateVolume(...)` logic to respect the [idempotency requirements](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume) described in the CSI spec. This change should hopefully make volume creation more reliable, especially for long-running and error-prone operations like large volume cloning.

#### E2E Testing

Not currently implemented, see #140.

#### Manual Testing

NOTE: These steps would previously fail because a pre-existing volume would not be respected by CreateVolume(...) after a timeout error.

1. Build the OCI image and push it to an image registry:
```
export REGISTRY_NAME=index.docker.io/mydockerhubuser
export IMAGE_VERSION=dev
```

2. Provision a Kubernetes cluster on Linode. (e.g. RKE2)
3. Navigate to `helm-chart/csi-driver`.
4. Update values.yaml to point to the image pushed in step 1.
5. Deploy the Helm chart to your Kubernetes cluster:
```
export KUBECONFIG=mykubeconfig.yaml
helm install linode-bs-csi .
```
6. Apply a `linode-block-storage` PVC and a job to populate it with random data:
```bash
kubectl apply -f - <<EOF
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-1
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 100Gi
  storageClassName: linode-block-storage
---
apiVersion: batch/v1
kind: Job
metadata:
  name: pvc-populator
spec:
  template:
    spec:
      containers:
        - name: populate
          image: alpine:latest
          command: ["dd", "if=/dev/urandom", "of=/mnt/pvc/foo", "bs=1M", "count=75000"]
          volumeMounts:
            - mountPath: /mnt/pvc
              name: data
      volumes:
        - name: data
          persistentVolumeClaim:
            claimName: pvc-1
      restartPolicy: Never
EOF
```
7. Wait for the job to complete.
```bash
kubectl wait --for=condition=complete job/pvc-populator 
```
8. Create a new PVC cloned from the first PVC (this should take several minutes to finish)
```bash
kubectl apply -f - <<EOF
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-2
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 100Gi
  storageClassName: linode-block-storage
  dataSource:
    kind: PersistentVolumeClaim
    name: pvc-1
EOF
```
9. Once the corresponding Linode volume has finished provisioning, observe that the PVC is bound in Kubernetes.
10. Ensure that only one Linode volume clone event has occurred.
11. (Optional) Check the CSI controller logs to see the volume creation fail repeatedly until it eventually reconciles.

### DIFF
--- a/pkg/linode-bs/controllerserver.go
+++ b/pkg/linode-bs/controllerserver.go
@@ -601,7 +601,7 @@ func (linodeCS *LinodeControllerServer) createLinodeVolume(
 	if err != nil {
 		return nil, status.Errorf(
 			codes.Internal,
-			"failed to create linode volume: %w", err,
+			"failed to create linode volume: %s", err,
 		)
 	}
 
@@ -619,7 +619,7 @@ func (linodeCS *LinodeControllerServer) cloneLinodeVolume(
 	if err != nil {
 		return nil, status.Errorf(
 			codes.Internal,
-			"failed to clone linode volume %d into new volume: %w", sourceID, err,
+			"failed to clone linode volume %d into new volume: %s", sourceID, err,
 		)
 	}
 
--- a/pkg/linode-bs/controllerserver.go
+++ b/pkg/linode-bs/controllerserver.go
@@ -76,14 +76,6 @@ func (linodeCS *LinodeControllerServer) CreateVolume(ctx context.Context, req *c
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	// Attempt to get info about the source volume.
-	// sourceVolumeInfo will be null if no content source is defined.
-	contentSource := req.GetVolumeContentSource()
-	sourceVolumeInfo, err := linodeCS.attemptGetContentSourceVolume(ctx, contentSource)
-	if err != nil {
-		return nil, err
-	}
-
 	// to avoid mangled requests for existing volumes with hyphen,
 	// we only strip them out on creation when k8s invented the name
 	// this is still problematic because we strip "-" from volume-name-prefixes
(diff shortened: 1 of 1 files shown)


pkg/linode-bs/controllerserver.go
```diff
@@ -120,47 +102,48 @@ func (linodeCS *LinodeControllerServer) CreateVolume(ctx context.Context, req *c
...
+		})
+
+		if err := linodeCS.CloudProvider.ResizeVolume(ctx, vol.ID, targetSizeGB); err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to resize cloned volume (%d): %s", targetSizeGB, err)
+		}
+	}
+
+	// Always wait for the volume to be active
+	if _, err := linodeCS.CloudProvider.WaitForVolumeStatus(
+		ctx, vol.ID, linodego.VolumeActive, waitTimeout); err != nil {
```
lgarber-akamai, line 145: Would it make sense to use a separate timeout const for clone operations? I'm hesitant because clone runtime can vary drastically depending on the contents of the disk (?).

lgarber-akamai: >Seems to make a lot of sense to use a different timeout for clone as I'd imagine (hope) a typical create volume operation would be much faster than cloning a multi GiB volume.

Sounds good, I'll implement that when I find some time.

>What happens in the case that the timeout gets hit - we return an error but the operation continues in the Linode API? Is there some kind of reconciliation elsewhere that'd account for the operation finishing?

So if the timeout is hit while [polling for volume readiness](https://github.com/linode/linode-blockstorage-csi-driver/pull/142/files#diff-a45fd181b265bc173ff27928709318c12e64ee8eedb87832081ef107dbc96197R144-R147), an error will be returned but the volume will **not** be destroyed. This error will implicitly trigger the volume creation request to be re-run, in which case [this bit of logic](https://github.com/linode/linode-blockstorage-csi-driver/pull/142/files#diff-a45fd181b265bc173ff27928709318c12e64ee8eedb87832081ef107dbc96197R560-R578) will discover the volume created during the previous run and will continue polling for readiness.

avestuk (approved):

PR #147, opened 2023-12-15
Improve readability of test errors; update dependencies; improve test invocation

okokes-akamai:
I started off just improving our newly working e2e tests, but ended up doing a few extra maint things I found useful:

- updated Expect/Eventually to make `BeTrue` failures more readable
- rewrote the `WaitForReady` retry loop to avoid the `appscode/go` dependency (redundant + long abandoned) -> removed the dependency as a result
- bumped ginkgo (and gomega) for better test progress reporting
- increased verbosity of tests, removed the deprecated `--progress` flag
- added parallelism to tests (4m -> 90s in terms of runtime)
- did _not_ test against v0.6.0, only against a random LKE cluster (v0.5.2 I believe), so I had to `--skip=Raw` for it to work

Note: I'm not a huge fan of `func (...) IsPodReady(...) error` as the naming suggests a bool return value. I did try that at first, but found the bool to be redundant, since I could easily encode the message in an error, but didn't find a more suitable name for the method.


avestuk (approved):

luthermonson (approved):

luthermonson:
```
Ran 5 of 5 Specs in 80.387 seconds
SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 0 Skipped
```

PR #152, opened 2024-01-11
LKE-3931: Set maxVolumesPerNode to 7

### LINKED ISSUE #154: Linodes currently can only mount 7 volumes PVCs, not 8
## General:

----

## Feature Requests:
* [x] Have you explained your rationale for why this feature is needed? 
* [x] Have you offered a proposed implementation/solution? 

----

## Bug Reporting

Linodes need a boot disk, hence maximum PVCs is currently 7.

### Expected Behavior

Kubernetes scheduler should not try to assign more than 7 volumes to a single Linode.

### Actual Behavior

CSI node advertises that it can bind 8 volumes. This eventually results in pod failing to run when it tries to bind 8th volume on the node.

### Steps to Reproduce the Problem

It's not deterministic, but the following is likely to reproduce:

  1. Prepare 3 nodes with no bound PVCs.
  1. Create a StatefulSet, single replica, 7 volumes. This pod should run.
  1. Create another StatefulSet, many replicas, each single volume. Try to bind 21 to 24 volumes in total.
  1. Eventually, a pod will be scheduled on a node trying to bind 8th volume and it will fail to run.

#### Additional Notes

----

For general help or discussion, join the [Kubernetes Slack team](https://kubernetes.slack.com/messages/CD4B15LUR/details/) channel `#linode`. To sign up, use the [Kubernetes Slack inviter](http://slack.kubernetes.io/).

The [Linode Community](https://www.linode.com/community/questions/) is a great place to get additional support.

rsienko:
This is a quick fix for the default case when we have 8 slots and one slot is already taken by boot disk. There may be more slots in use by non-PVC volumes such as swap but they are not the default in LKE.

fixes #154

### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 

### DIFF
--- a/pkg/linode-bs/nodeserver.go
+++ b/pkg/linode-bs/nodeserver.go
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	maxVolumesPerNode = 8
+	maxVolumesPerNode = 7
 )
 
 type LinodeNodeServer struct {


luthermonson:
@kitknox i think you should weigh in on this, there is some logic he suggested we could use to check the disks on the node and make this limit more dynamic. this is probably the quick fix though and we could go with this and ticket the better fix if its too much of a time sink today

kitknox:
In the current situation where the API attempts to enforce a limit of 8 block devices, this is an improvement.  The downside is that work is planned to improve that to align with the host enforced scalable limit at which point this will be worse.  Given workloads are already impacted by the current oversubscription issue there is value in getting this out there but we need to ensure we come back to this quickly when the upstream API shifts.

apilny-akamai (approved):

avestuk (approved):

PR #163, opened 2024-04-04
Replace Terraform to CAPL under e2e tests

mhmxs:
Terraform provider of Linode has been deprecated. This change replaces Terraform to Linode CAPI provider.

Fixes: https://github.com/linode/linode-blockstorage-csi-driver/issues/151

### General:

### Pull Request Guidelines:

1. [X] Does your submission pass tests?
1. [ ] Have you added tests? 
1. [X] Are your commits atomic, addressing one change per commit?
1. [X] Are you following the conventions of the language? 
1. [X] Have you saved your large formatting changes for a different PR, so we can focus on your work?
1. [X] Have you explained your rationale for why this feature is needed? 
1. [X] Have you linked your PR to an [open issue](https://blog.github.com/2013-05-14-closing-issues-via-pull-requests/)


e2e/test/e2e_suite_test.go
```diff
@@ -28,7 +28,7 @@ var (
 	clusterName string
 	linodeDebug = false
 	linodeURL   = "https://api.linode.com"
-	region      = "us-east"
+	region      = "us-sea"
```
mhmxs, line 31: The new way needs a region where Metadata service is available. Does this change hurt?

e2e/test/scripts/infrastructure-linode/0.1.0/infrastructure-components.yaml
```diff
@@ -0,0 +1,1749 @@
...
+        - /manager
+        env:
+        - name: LINODE_TOKEN
+          valueFrom:
+            secretKeyRef:
+              key: apiToken
+              name: capl-manager-credentials
+        - name: LINODE_API_VERSION
+          value: v4beta
+        image: docker.io/linode/cluster-api-provider-linode:main
```
mhmxs, line 1697: Linode CAPI provider doesn't have any official releases, so this image doesn't exist! This PR has to wait until the provider has been released.

e2e/test/e2e_suite_test.go
lgarber-akamai: Should be fine, we've moved away from legacy sites for E2E testing in other projects for the most part

e2e/test/scripts/infrastructure-linode/0.1.0/clusterclass-kubeadm.yaml
```diff
@@ -0,0 +1,397 @@
...
+          secret:
+            key: sysctl-k8s.conf
+            name: common-init-files
+        path: /etc/sysctl.d/k8s.conf
+      - contentFrom:
+          secret:
+            key: kubeadm-pre-init.sh
+            name: common-init-files
+        path: /kubeadm-pre-init.sh
+        permissions: "0500"
```
avestuk, line 148: Don't need execute perms?

mhmxs: @avestuk the manifests were generated by the project itself. So nothing to do here.

e2e/test/scripts/create_cluster.sh
```diff
@@ -5,54 +5,79 @@ set -o pipefail
...
+
+clusterctl generate cluster ${CLUSTER_NAME} \
+  --flavor clusterclass-kubeadm \
+  --config $(realpath "$(dirname "$0")/clusterctl.yaml") \
+  | kubectl apply --wait -f -
+
+c=8
+until kubectl get secret ${CLUSTER_NAME}-kubeconfig; do
+  sleep $(((c--)))
+done
```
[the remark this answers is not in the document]
avestuk: #justbashthings

e2e/test/scripts/delete_cluster.sh
```diff
@@ -5,20 +5,9 @@ set -o pipefail
...
-if [[ -d "terraform.tfstate.d" ]]
-then
-    rm -rf terraform.tfstate.d
-fi
+export CLUSTER_NAME="$1"
+export KUBECONFIG="$(pwd)/${CLUSTER_NAME}-management.conf"
 
+timeout 5m kubectl delete linodemachine --all --wait
+timeout 5m kubectl delete linodecluster --all --wait
+timeout 5m kind delete cluster -n ${CLUSTER_NAME}
```
lgarber-akamai, line 13: nitpick: Should the local cluster deletion be handled through ctlptl rather than directly with `kind`?

e2e/Makefile
```diff
@@ -26,8 +26,10 @@ check-token:
 	  exit 1; \
 	fi
 
-install-terraform:
-	sudo apt-get install wget unzip
-	wget https://releases.hashicorp.com/terraform/0.11.13/terraform_0.11.13_linux_amd64.zip
-	unzip terraform_0.11.13_linux_amd64.zip
-	sudo mv terraform /usr/local/bin/
+install-dependencies:
```
luthermonson, line 29: i really liked how we got the CAPL makefile looking with the individual targets, goarch, version pins and go installs. since we're changing this any way can we move to that style?

e2e/README.md
```diff
@@ -5,7 +5,8 @@
 Install the following packages (macOS examples)
 
 ```
-brew install terraform
```
luthermonson, line 8: if we migrate to the CAPL style makefile could we also drop brew docs?

e2e/test/scripts/infrastructure-linode/0.1.0/cluster-template-clusterclass-kubeadm.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: cluster.x-k8s.io/v1beta1
```
eljohnson92, line 1: These files in `infrastructure-linode/*` should all be available as release artifacts once we have a release(hopefully this week)

deploy/kubernetes/base/csi-driver-instance.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: storage.k8s.io/v1beta1
+apiVersion: storage.k8s.io/v1
```
mhmxs, line 1: v1beta1 is not supported on modern Kind clusters.

e2e/Makefile
```diff
@@ -1,20 +1,25 @@
...
 
-test: ${GINKGO_PATH} check-token test-manifest
+clean-cluster: ctlptl
+	ctlptl delete cluster kind-management --ignore-not-found
+	rm -rf ./test/kind-management.conf
+
+test: clusterctl ginkgo check-token test-manifest test-cluster
 	go list -m; \
-	ginkgo -r --vv --procs=6 --trace --cover $(TEST_ARGS) -- --v=3 --k8s-version=${K8S_VERSION} $(SUITE_ARGS)
+	ginkgo -r --vv --procs=1 --trace --cover $(TEST_ARGS) -- --v=3 --k8s-version=${K8S_VERSION} $(SUITE_ARGS)
```
mhmxs, line 20: There is a limitation on Linode API, that's why only `procs=1` supported at the moment.

deploy/kubernetes/base/csi-driver-instance.yaml
avestuk: This API was removed in 1.27 and the v1 API has been available since 1.24 so I think this change is ok to make.

e2e/test/scripts/create_cluster.sh
```diff
@@ -5,54 +5,51 @@ set -o pipefail
...
-EOF
-
-terraform workspace new ${CLUSTER_NAME}
-
-terraform init
-
-terraform apply \
- -auto-approve
+METADATA_REGIONS="nl-ams in-maa us-ord id-cgk us-lax es-mad us-mia it-mil jp-osa fr-par br-gru us-sea se-sto us-iad"
+[[ "$METADATA_REGIONS" =~ .*"${LINODE_REGION}".* ]] || (echo "Given region doesn't support Metadata service" ; exit 1)
```
avestuk, line 25: I believe CAPL supports regions without metadata now. @eljohnson92 can you confirm?

avestuk (approved):

avestuk (commented):
The cluster provisioning works (I had to make a few tweaks I'll raise a PR) but running `make test` results in docker.io/linode/linode-blockstorage-csi-driver:canary being deployed which fails to start with a `exec /linode: exec format error`

It's not a blocker but we should improve the documentation to explain how somebody can easily test an image built from their local machine. I'll be glad to raise a PR for this.

mhmxs:
> docker.io/linode/linode-blockstorage-csi-driver:canary

I had to replace it to the latest stable release. I think this would be a temporary building issue around the canary.

e2e/Makefile
avestuk: I've been running tests today with procs=2 and it takes 50% less time and hasn't showed any issues. wdyt?

mhmxs: @avestuk could you please change it on your polishing PR?

e2e/test/scripts/create_cluster.sh
luthermonson: it does now with a stackscript shim

mhmxs: @luthermonson do we want to support it?

e2e/Makefile
avestuk: Yes I already have :D

e2e/test/scripts/create_cluster.sh
luthermonson: The way I understand it we shouldn't need the region check and CAPL will just work in any region when we pin and use the official release. Probably safe to remove this check and rely on CAPL completely but I'll let @eljohnson92 make the call since he wrote the feature.

e2e/test/scripts/create_cluster.sh
```diff
@@ -5,54 +5,51 @@ set -o pipefail
 set -o nounset
 set -x
 
-export LINODE_API_TOKEN="$1"
+export LINODE_TOKEN="$1"
 export CLUSTER_NAME="$2"
-export K8S_VERSION="$3"
+export KUBERNETES_VERSION="$3"
+export CAPLI_VERSION="0.1.0"
```
eljohnson92, line 11: We have landed on CAPL for the naming instead of CAPLI, could we rename these references to CAPL?

e2e/Makefile
```diff
@@ -1,20 +1,25 @@
...
-.PHONY: test clean check-token install-terraform test-manifest
+.PHONY: test clean check-token test-manifest
 
-test-manifest:
+test-manifest: kustomize
 	../hack/generate-yaml.sh $(IMAGE_VERSION) > test/manifest/linode-blockstorage-csi-driver.yaml
 
-${GINKGO_PATH}:
-	go install github.com/onsi/ginkgo/v2/ginkgo@v2.13.2
+test-cluster: ctlptl clean-cluster
```
lgarber-akamai, line 11: Would it make sense to make `test-manifest` a dependency of `test-cluster` since `linode-blockstorage-csi-driver.yaml` is consumed [here](https://github.com/linode/linode-blockstorage-csi-driver/pull/163/files#diff-0908e8aad04dc51bce3a028b466879848a0cf5e57f2ee1af8bf08a599c600e23R24)?

lgarber-akamai (approved):
I'm seeing some weird issues with the E2E test suite itself but they seem unrelated to the testing infrastructure changes. LGTM!

e2e/test/scripts/infrastructure-linode/0.1.0/infrastructure-components.yaml
eljohnson92: CAPL now has it's first release, you can use the `:latest` image tag

e2e/test/scripts/infrastructure-linode/0.1.0/cluster-template-clusterclass-kubeadm.yaml
eljohnson92: CAPL has it's first release so these should be able to be removed at this point

e2e/test/scripts/create_cluster.sh
eljohnson92: that is correct, CAPL is able to detect if it needs to use the stackscript shim or not automatically and should just work transparently in all regions

e2e/test/scripts/create_cluster.sh
```diff
@@ -5,54 +5,48 @@ set -o pipefail
...
+  --flavor clusterclass-kubeadm \
+  --config clusterctl.yaml \
+  | kubectl apply --wait -f -)
 
+c=8
+until kubectl get secret -n ${CLUSTER_NAME} ${CLUSTER_NAME}-kubeconfig; do
+  sleep $(((c--)))
+done
+
+kubectl get secret -n ${CLUSTER_NAME} ${CLUSTER_NAME}-kubeconfig -o jsonpath="{.data.value}" \
```
eljohnson92, line 36: `clusterctl get kubeconfig ${CLUSTER_NAME} > "$(pwd)/${CLUSTER_NAME}.conf"` should get the kubeconfig for you if you'd rather use clusterctl instead of the kubectl get secret

eljohnson92 (approved):

PR #159, opened 2024-03-11
Write encryptionKey to crypsetup stdin

avestuk:
### General:

* [x] Have you removed all sensitive information, including but not limited to access keys and passwords?
* [x] Have you checked to ensure there aren't other open or closed [Pull Requests](../../pulls) for the same bug/feature/question?

### Pull Request Guidelines:

1. [ ] Does your submission pass tests?
1. [ ] Have you added tests? 
1. [x] Are you addressing a single feature in this PR? 
1. [x] Are your commits atomic, addressing one change per commit?
1. [x] Are you following the conventions of the language? 
1. [x] Have you saved your large formatting changes for a different PR, so we can focus on your work?
1. [x] Have you explained your rationale for why this feature is needed? 
1. [x] Have you linked your PR to an [open issue](https://blog.github.com/2013-05-14-closing-issues-via-pull-requests/)

Provides a fix for #158 by providing the encryption key on stdin rather than writing it to disk. This eliminates the requirement for `/tmp` to be a tmpfs as we just don't write the key to disk at all. 

I've also made a few other changes to make the code a little more idiomatic. 

Adding unit tests is doable - cryptsetup will create a loopback device if [we target a file instead of a block device](https://man7.org/linux/man-pages/man8/cryptsetup.8.html), we'd need to make further changes in order to make the functions unit testable. I'll raise a separate issue for doing this. 

Can be tested using following manifests and the container ghcr.io/avestuk/linode-blockstorage-csi-driver:helm-v0.6.3-5-g763d4d2

```
apiVersion: v1
data:
[...]

PR #165, opened 2024-04-10
Luks e2e

avestuk:
### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 

Addresses the lack of tests in #159 - Creating loopback devices in docker was sadly not as straight forward as I had hoped and I wasn't able to get something working. Given the ease with which we can now write e2e tests thanks to #163 and @mhmxs an e2e test should afford us a better level of comfort. 

Currently this PR is set as a draft to get some initial commentary back on the overall approach. Mainly I'm not enamoured with using static names for the LUKS storageClass and Secrets and the subsequent cleanup that has to happen. I don't think that there is an issue in modifying the "global" `storageClass` as ginkgo does parallelism using separate processes so mutation of `storageClass` in one process will not affect others. 

Breaking this out into a separate `Describe` node would work but would require copying `JustBeforeEach` or moving it to a function and substantially altering the tests. I would like to add additional LUKS tests to show that LUKS volumes require a proper AES-XTS key in order to function, there may well be other tests we would like to add such as LUKS volume expansion etc.

### DIFF
--- a/e2e/test/csi_driver_test.go
+++ b/e2e/test/csi_driver_test.go
@@ -1,10 +1,11 @@
 package test
 
 import (
-	"e2e_test/test/framework"
 	"fmt"
 	"math/rand"
 	"strconv"
+
+	"e2e_test/test/framework"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
--- a/e2e/test/csi_driver_test.go
+++ b/e2e/test/csi_driver_test.go
@@ -308,8 +308,58 @@ var _ = Describe("Linode CSI Driver", func() {
 					volumeType = core.PersistentVolumeBlock
 					size = "10Gi"
 				})
-				It("should write and read", func() {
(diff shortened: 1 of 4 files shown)


avestuk:
@nesv @eljohnson92  Hello from the other side 👋 Tagging the two of you as you've been here recently. 

Do you want to take this PR over? It's just an E2E test for the LUKS encrypted volumes, at least one user was interested in LUKS encrypted volumes because they can manage their own keys and having an e2e test seems to make sense. The reason that I statically defined the LUKS key was that I never quite figured out how the keys should be generated. You'd think it'd just be 256/512 bytes of random stuff but it didn't quite seem to work for me. 

Anyway if you want to throw this in the bin then by all means go ahead :)

eljohnson92:
Hey @avestuk thanks for the context here! Yes, we will take this effort over. We have a separate initiative to revamp the e2e tests. We'll likely roll this in with those changes

PR #169, opened 2024-06-03
Remove "pkg/linode-bs/metadata"

nesv:
These changes drop `pkg/linode-bs/metadata` in favour of a `Metadata` struct in `pkg/linode-bs` that can be retrieved from either the `GetMetadata` or `GetMetadataFromAPI` functions in the same package. I'm sure there were good intentions around defining an interface for retrieving metadata for a node, but for how the metadata was being retrieved and used, and how it was faked out for testing, it felt a little overwrought.

This pull request also adds a small change to the `get-linode-id` script provided by the Helm chart (and invoked in an init container): the use of awk(1) is dropped for using the `-s, --string KEYWORD` flag to dmidecode(8).

### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 


pkg/linode-bs/driver_test.go
```diff
@@ -2,25 +2,22 @@ package linodebs
...
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	linodeclient "github.com/linode/linode-blockstorage-csi-driver/pkg/linode-client"
-	"github.com/linode/linode-blockstorage-csi-driver/pkg/metadata"
 	mountmanager "github.com/linode/linode-blockstorage-csi-driver/pkg/mount-manager"
 
-	"strconv"
```
AshleyDumaine, line 19: Not for this PR, but I'm wondering if we should consider using `gci` for linting import order like we did on CAPL (though it seems we don't actually even have a .golangci.yml on this repo? 🤔): https://github.com/linode/cluster-api-provider-linode/commit/158b033326eb5ffbea4b6cc82d3f5324f9aec247

nesv: Yeah, we definitely could!

Would you be willing to submit a PR with that change, or file an issue (or ticket) so that I don't forget to do it? :smile:

AshleyDumaine: I'll put it in a ticket 👍

AshleyDumaine (approved):

eljohnson92 (approved):

PR #179, opened 2024-06-24
Fix script to look at just the id

tchinmai7:
### General:

The get-linode-id script can produce a mangled linode-id when there's a tag with a `id` in there. For example with a metadata response like this
```
backups.enabled: true
backups.status: pending
host_uuid: 2d56ceffafec9f76be86e1d2d1dd37ec2204b28f
id: 60586846
label: jkvbpzyzqaa9pj8ix-master-0
region: us-ord
specs.disk: 163840
specs.gpus: 0
specs.memory: 8192
specs.transfer: 5000
specs.vcpus: 4
tags: cluster:kpp-api-single
tags: clusterUid:jkvbpzyzqaa9pj8ix
tags: prov-from:mgmt-cluster.akamai-kpp.com
type: g6-standard-4
```
the script ends up with
```
60586846clusterUid:jkvbpzyzqaa9pj8i
```

This fixes it by explictly asking for json and looking only at the "id" param.

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/helm-chart/csi-driver/templates/get-linode-id.yaml
+++ b/helm-chart/csi-driver/templates/get-linode-id.yaml
@@ -23,7 +23,7 @@ data:
       # Try to get ID using Metadata API with a 5-second timeout
       TOKEN=$(curl -sS -m 5 -X PUT -H "Metadata-Token-Expiry-Seconds: 3600" $META_API/token || true)
       if [ ! -z "$TOKEN" ]; then
-        ID=$(curl -sS -m 5 -H "Metadata-Token: $TOKEN" $META_API/instance -H "Accept: application/json" | grep -o '"id": [0-9]*' | cut -d' ' -f2 |  tr -d '\n' || true)
+        ID=$(curl -sS -m 5 -H "Metadata-Token: $TOKEN" $META_API/instance -H "Accept: application/json" | jq -r .id |  tr -d '\n' || true)
         echo "Metadata linode ID was $ID"
       fi
     fi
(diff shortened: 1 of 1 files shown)


ericsund (approved):

nesv (approved):

helm-chart/csi-driver/templates/get-linode-id.yaml
```diff
@@ -23,7 +23,7 @@ data:
       # Try to get ID using Metadata API with a 5-second timeout
       TOKEN=$(curl -sS -m 5 -X PUT -H "Metadata-Token-Expiry-Seconds: 3600" $META_API/token || true)
       if [ ! -z "$TOKEN" ]; then
-        ID=$(curl -sS -m 5 -H "Metadata-Token: $TOKEN" $META_API/instance -H "Accept: application/json" | grep -o '"id": [0-9]*' | cut -d' ' -f2 |  tr -d '\n' || true)
+        ID=$(curl -sS -m 5 -H "Metadata-Token: $TOKEN" $META_API/instance -H "Accept: application/json" | jq -r .id |  tr -d '\n' || true)
```
luthermonson, line 26: context... where does this run and do we for sure have JQ there?

tchinmai7: this runs in the init container - which defaults to `alpine/k8s` https://hub.docker.com/r/alpine/k8s - which does have jq

luthermonson (approved):

helm-chart/csi-driver/templates/get-linode-id.yaml
tchinmai7: confirmed it exists 


```
$ docker run -it alpine/k8s:1.25.14 bash
dee1d4fe298c:/apps# which jq
/usr/bin/jq
```

PR #184, opened 2024-07-03
internal/driver: Fix off-by-one error with max volume attachments

### LINKED ISSUE #182: Linodes cannot mount more than 7 PVC on instances with 8GB or less.
## General:

----

## Feature Requests:
* [ ] Have you explained your rationale for why this feature is needed? 
* [ ] Have you offered a proposed implementation/solution? 

----

## Bug Reporting

Small Linodes (i tested 4GB and 8GB) cannot have more than 7 attached PVC per node. According to the [documentation](https://www.linode.com/docs/products/storage/block-storage/#limits-and-considerations) local disks and block storage are counted aganist the max volumes that can be attached to a node.

### Expected Behavior

The pod should be relocated to another node since the max number of PVC has been reached.

### Actual Behavior

This gets emmited non-stop on the pod trying to attach a 8th volume to the node. The pod remains on the node.

```
AttachVolume.Attach failed for volume "pvc-xxxxxxxxxxx" : rpc error: code = ResourceExhausted desc = max number of volumes (8) already attached to instance
```

### Steps to Reproduce the Problem

1. Prepare 3 nodes with no bound PVCs.
1. Create a StatefulSet, single replica, 7 volumes. This pod should run.
1. Create another StatefulSet, many replicas, each single volume. Try to bind 21 to 24 volumes in total.
1. Eventually, a pod will be scheduled on a node trying to bind 8th volume and it will fail to run.

#### Additional Notes

Related to #154 and probably reintroduced in v0.7.0. Haven't tested but it could also apply to bigger Linodes that allows more volumes but counting attached volumes incorrectly.

----

For general help or discussion, join the [Kubernetes Slack team](https://kubernetes.slack.com/messages/CD4B15LUR/details/) channel `#linode`. To sign up, use the [Kubernetes Slack inviter](http://slack.kubernetes.io/).

The [Linode Community](https://www.linode.com/community/questions/) is a great place to get additional support.

nesv:
When calculating the maximum number of allowed volume attachments, the code was previously taking the ideal maximum number of volumes that could be attached to a node. The way the attachment was calculated, it treated instance disks the same as volumes, which is not correct.

This commit fixes what is effectively an off-by-one error, by subtracting the number of instance disks from the theoretical maximum number of block devices that can be attached to the instance.

In other words, controller and node servers will now report the number of block storage volumes that can be attached, not just block devices.

Fixes #182

### DIFF
--- a/internal/driver/controllerserver.go
+++ b/internal/driver/controllerserver.go
@@ -265,10 +265,19 @@ func (linodeCS *LinodeControllerServer) ControllerPublishVolume(ctx context.Cont
 	}
 
 	// Check to see if there is room to attach this volume to the instance.
-	if canAttach, err := linodeCS.canAttach(ctx, instance); err != nil {
+	if canAttach, err := linodeCS.canAttach(ctx, instance); errors.Is(err, errNilInstance) {
+		return &csi.ControllerPublishVolumeResponse{}, status.Error(codes.Internal, "cannot determine volume attachments for a nil instance")
+	} else if err != nil {
 		return &csi.ControllerPublishVolumeResponse{}, status.Error(codes.Internal, err.Error())
 	} else if !canAttach {
-		limit := linodeCS.maxVolumeAttachments(instance)
+		// If we can, try and add a little more information to the error message
+		// for the caller.
+		limit, err := linodeCS.maxVolumeAttachments(ctx, instance)
+		if errors.Is(err, errNilInstance) {
+			return &csi.ControllerPublishVolumeResponse{}, status.Error(codes.Internal, "cannot calculate max volume attachments for a nil instance")
+		} else if err != nil {
+			return &csi.ControllerPublishVolumeResponse{}, status.Error(codes.ResourceExhausted, "max number of volumes already attached to instance")
+		}
 		return &csi.ControllerPublishVolumeResponse{}, status.Errorf(codes.ResourceExhausted, "max number of volumes (%d) already attached to instance", limit)
 	}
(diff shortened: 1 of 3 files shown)


nesv:
Squashed fixup commits.

internal/driver/controllerserver.go
```diff
@@ -268,7 +268,10 @@ func (linodeCS *LinodeControllerServer) ControllerPublishVolume(ctx context.Cont
 	if canAttach, err := linodeCS.canAttach(ctx, instance); err != nil {
 		return &csi.ControllerPublishVolumeResponse{}, status.Error(codes.Internal, err.Error())
 	} else if !canAttach {
-		limit := linodeCS.maxVolumeAttachments(instance)
+		limit, err := linodeCS.maxVolumeAttachments(ctx, instance)
+		if err != nil {
+			return &csi.ControllerPublishVolumeResponse{}, status.Error(codes.ResourceExhausted, "max number of volumes already attached to instance")
```
AshleyDumaine, line 273: not sure this error is right (e.g. instance being nil or an api error listing the instance disks)

nesv: Ah, good catch! Thank you. I will handle those errors separately.

AshleyDumaine (approved):
LGTM

nesv:
Squashed fixup commits (again).

rahulait (approved):
lgtm

eljohnson92 (approved):

PR #191, opened 2024-07-17
[Clean Up] - NodeStageVolume, NodeUnStageVolume, Adding golangci lint to GHA, and various small cleans

komer3:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


pkg/common/idhelpers.go
```diff
@@ -20,7 +20,8 @@ type (
 	}
 )
 
-const linodeBSLabelLength = 32
+// TODO: Rename this variable
+const linodeStorageLabelLength = 32
```
AshleyDumaine, line 24: might want to change this and other variables from storage to volume?

komer3: Oh yup linodeVolumeLabelLenght is more fitting!

shanduur (approved):

eljohnson92:
it looks like you need remove the `./mock` directory from code coverage reporting here

devbox.json
```diff
@@ -1,6 +1,6 @@
 {
   "packages": [
-    "go_1_22@latest",
```
eljohnson92, line 3: I think we should keep this pinned to 1.22 if we can as upgrades past 1.22 will likely need some more intentional work to do

internal/driver/luks.go
```diff
@@ -122,20 +132,23 @@ func luksFormat(ctx LuksContext, source string) error {
...
-	if err != nil {
-		return fmt.Errorf("failed to get stdin pipe for cryptsetup, got err: %w", err)
-	}
+	// Set the encryption key as stdin
+	cmd.SetStdin(strings.NewReader(ctx.EncryptionKey))
 
-	if _, err := io.WriteString(stdin, ctx.EncryptionKey); err != nil {
-		return fmt.Errorf("failed to write to stdin pipe for cryptsetup, got err: %w, closing pipe: %w", err, stdin.Close())
-	}
+	// stdin, err := cmd.StdinPipe()
```
eljohnson92, line 140: we should remove commented-out code here or describe why it needs to stay commented out, a comment describing what functionality we might want to add in later is likely a better approach

komer3: I'm leaving it here until Luks E2E tests are merged in. Its to make sure the new code works and didn't break luks functionality.

internal/driver/nodeserver.go
```diff
@@ -236,53 +234,16 @@ func (ns *LinodeNodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeSt
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
 
-	// VolumeMode=block
+	// Check if the volume mode is set to 'Block'
 	// Do nothing else with the mount point for stage
-	if blk := volumeCapability.GetBlock(); blk != nil {
+	if blk := req.VolumeCapability.GetBlock(); blk != nil {
```
bcm820, line 239: Do we need a `nil` check on `req.VolumeCapability` here since the original implementation had one?

komer3: Yup. I'm trying to keep the original code the same and not make any changes unless absolutely needed

bcm820: Oh, I was not referring to the return value of `GetBlock`, but `req.VolumeCapability` itself. I just noticed that there was a nil check on it before but I don't see it being done here.

bcm820 (approved):

eljohnson92 (approved):

PR #196, opened 2024-08-13
[fix] Address the errors we are seeing for NodeExpandVolume() in CSI-Sanity test

komer3:
The cases being addressed here from csi-sanity tests:
- Node Service NodeExpandVolume should fail when no volume id is provided
- Node Service NodeExpandVolume should fail when no volume path is provided
- Node Service NodeExpandVolume should fail when volume is not found
- Node Service NodeExpandVolume should work if node-expand is called after node-publish

**NOTE:** This PR is not for implementing the `NodeExpandVolume()`. That method is in the roadmap for future PRs. This is just meant for address errors related to validation of the request object.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/internal/driver/nodeserver.go
+++ b/internal/driver/nodeserver.go
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -25,6 +26,7 @@ import (
 	"github.com/linode/linode-blockstorage-csi-driver/pkg/common"
 	linodeclient "github.com/linode/linode-blockstorage-csi-driver/pkg/linode-client"
 	mountmanager "github.com/linode/linode-blockstorage-csi-driver/pkg/mount-manager"
+	"github.com/linode/linodego"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -275,6 +277,28 @@ func (ns *LinodeNodeServer) NodeUnstageVolume(ctx context.Context, req *csi.Node
 
 func (ns *LinodeNodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
 	klog.V(4).Infof("NodeExpandVolume called with req: %#v", req)
+
+	// Validate req (NodeExpandVolumeRequest)
+	err := validateNodeExpandVolumeRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check linode to see if a give volume exists by volume ID
+	// Make call to linode api using the lindoe api client
+	LinodeVolumeKey, err := common.ParseLinodeVolumeKey(req.GetVolumeId())
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "Error parsing linode volume key: %v", err)
+	}
+	jsonFilter, err := json.Marshal(map[string]string{"label": LinodeVolumeKey.Label})
+	if err != nil {
+		return nil, errInternal("marshal json filter: %v", err)
+	}
+	_, err = ns.CloudProvider.ListVolumes(ctx, linodego.NewListOptions(0, string(jsonFilter)))
+	if err != nil {
+		fmt.Printf("list volumes: %v\n", err)
+		return nil, status.Errorf(codes.NotFound, "list volumes: %v", err)
+	}
 
 	return &csi.NodeExpandVolumeResponse{
 		CapacityBytes: req.CapacityRange.RequiredBytes,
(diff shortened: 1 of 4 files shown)


internal/driver/nodeserver.go
```diff
@@ -276,6 +278,28 @@ func (ns *LinodeNodeServer) NodeUnstageVolume(ctx context.Context, req *csi.Node
 func (ns *LinodeNodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
 	klog.V(4).Infof("NodeExpandVolume called with req: %#v", req)
 
+	// Validate req (NodeExpandVolumeRequest)
+	err := validateNodeExpandVolumeRequest(req)
+	if err != nil {
```
AshleyDumaine, line 283: nit: 
```suggestion
	if err := validateNodeExpandVolumeRequest(req); err != nil {
```

internal/driver/nodeserver.go
```diff
@@ -276,6 +278,28 @@ func (ns *LinodeNodeServer) NodeUnstageVolume(ctx context.Context, req *csi.Node
...
+		return nil, status.Errorf(codes.NotFound, "Error parsing linode volume key: %v", err)
+	}
+	jsonFilter, err := json.Marshal(map[string]string{"label": LinodeVolumeKey.Label})
+	if err != nil {
+		return nil, errInternal("marshal json filter: %v", err)
+	}
+	_, err = ns.CloudProvider.ListVolumes(ctx, linodego.NewListOptions(0, string(jsonFilter)))
+	if err != nil {
+		fmt.Printf("list volumes: %v\n", err)
```
AshleyDumaine, line 299: should this print statement still be here?

internal/driver/nodeserver.go
```diff
@@ -276,6 +278,28 @@ func (ns *LinodeNodeServer) NodeUnstageVolume(ctx context.Context, req *csi.Node
...
+	LinodeVolumeKey, err := common.ParseLinodeVolumeKey(req.GetVolumeId())
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "Error parsing linode volume key: %v", err)
+	}
+	jsonFilter, err := json.Marshal(map[string]string{"label": LinodeVolumeKey.Label})
+	if err != nil {
+		return nil, errInternal("marshal json filter: %v", err)
+	}
+	_, err = ns.CloudProvider.ListVolumes(ctx, linodego.NewListOptions(0, string(jsonFilter)))
+	if err != nil {
```
AshleyDumaine, line 298:
```suggestion
	if _, err = ns.CloudProvider.ListVolumes(ctx, linodego.NewListOptions(0, string(jsonFilter))); err != nil {
```

AshleyDumaine (commented):
few minor comments

internal/driver/nodeserver.go
komer3: ops good catch

AshleyDumaine (approved):

internal/driver/nodeserver.go
```diff
@@ -276,6 +278,25 @@ func (ns *LinodeNodeServer) NodeUnstageVolume(ctx context.Context, req *csi.Node
 func (ns *LinodeNodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
 	klog.V(4).Infof("NodeExpandVolume called with req: %#v", req)
 
+	// Validate req (NodeExpandVolumeRequest)
+	if err := validateNodeExpandVolumeRequest(req); err != nil {
+		return nil, err
+	}
+
+	// Check linode to see if a give volume exists by volume ID
+	// Make call to linode api using the lindoe api client
```
rahulait, line 287:
```suggestion
	// Make call to linode api using the linode api client
```

internal/driver/nodeserver.go
```diff
@@ -276,6 +278,25 @@ func (ns *LinodeNodeServer) NodeUnstageVolume(ctx context.Context, req *csi.Node
...
+	LinodeVolumeKey, err := common.ParseLinodeVolumeKey(req.GetVolumeId())
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "Error parsing linode volume key: %v", err)
+	}
+	jsonFilter, err := json.Marshal(map[string]string{"label": LinodeVolumeKey.Label})
+	if err != nil {
+		return nil, errInternal("marshal json filter: %v", err)
+	}
+	if _, err = ns.CloudProvider.ListVolumes(ctx, linodego.NewListOptions(0, string(jsonFilter))); err != nil {
```
rahulait, line 296: Not sure how many times this api call will be made (thinking just once or twice when we run the tests), if its very few times, then we are ok. If this call will be made periodically every few mins, then we need to discuss if we should introduce additional calls as we we don't do anything with such requests and simply return. I am hoping its just for tests, so this should be fine. Let me know if there are any other cases when this will be triggered.

komer3: The Api call isn't made very often. In this case, the test never really reaches it either.

rahulait (approved):
LGTM

PR #197, opened 2024-08-14
[feat] : add support to run upstream k8s e2e tests

rahulait:
### General:

This PR allows one to run upstream k8s e2e tests for storage class. We intend to run these on the main branch whenever any change is merged to main.

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
       LINODE_REGION: us-sea
       LINODE_CONTROL_PLANE_MACHINE_TYPE: g6-standard-2
       LINODE_MACHINE_TYPE: g6-standard-2
-      WORKER_NODES: ${{ github.head_ref == 'upstream-e2e-tests' && '3' || '0' }}
+      WORKER_NODES: ${{ github.ref == 'refs/heads/main' && '3' || '0' }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -104,7 +104,7 @@ jobs:
         run: devbox run csi-sanity-test
 
       - name: run upstream E2E Tests
-        if: github.head_ref == 'upstream-e2e-tests'
+        if: github.ref == 'refs/heads/main'
         run: devbox run upstream-e2e-tests
 
       - name: Cleanup Resources
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,7 @@ jobs:
       LINODE_REGION: us-sea
       LINODE_CONTROL_PLANE_MACHINE_TYPE: g6-standard-2
       LINODE_MACHINE_TYPE: g6-standard-2
+      WORKER_NODES: ${{ github.head_ref == 'upstream-e2e-tests' && '3' || '0' }}
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,10 @@ jobs:
       - name: Run CSI-Sanity Tests
         run: devbox run csi-sanity-test
 
+      - name: run upstream E2E Tests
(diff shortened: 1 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -102,6 +102,10 @@ jobs:
       - name: Run CSI-Sanity Tests
         run: devbox run csi-sanity-test
 
+      - name: run upstream E2E Tests
+        if: github.ref_head == 'upstream-e2e-tests'
```
rahulait, line 106: This will be changed to main before merging. Currently, its set to branch name just for testing purpose.

Makefile
```diff
@@ -93,6 +105,7 @@ remote-cluster-deploy:
 	clusterctl generate cluster $(TEST_CLUSTER_NAME) \
 		--kubernetes-version $(K8S_VERSION) \
 		--infrastructure linode-linode:$(CAPL_VERSION) \
+		--control-plane-machine-count 1 --worker-machine-count 1 \
```
komer3, line 108: Question: Any reason to create 2 nodes for testing? Is one node getting overload?

hack/upstream-e2e-tests.sh
```diff
@@ -0,0 +1,30 @@
...
+tar xzvf ${OUT_TAR} -C ${OUT_DIR}
+
+# Run k8s e2e tests for storage driver
+./${OUT_DIR}/kubernetes/test/bin/e2e.test \
+    -ginkgo.v \
+    -ginkgo.focus='External.Storage' \
+    --ginkgo.skip='disruptive' \
+    --ginkgo.skip='ephemeral' \
+    --ginkgo.skip='volume-expand' \
+    --ginkgo.skip='multiVolume' \
```
komer3, line 22: Question: What is the multiVolume testing for?

Makefile
rahulait: Pods were not getting scheduled on control-plane node due to taints. Hence, I provisioned a worker node so that pods can be scheduled there by the tests.

hack/upstream-e2e-tests.sh
rahulait: It runs tests for a pod having more than 1 volume attached and then tries to see how it behaves if the pod is killed and scheduled on same node again and when scheduled on different node. I was seeing some errors (not sure if they were bcoz I had only one schedulable worker node and it was trying to reschedule pod to different node) and hence had disabled it for now. Do you think we should enable this and try testing with a setup having more than one worker nodes?

.github/workflows/ci.yml
```diff
@@ -76,6 +76,7 @@ jobs:
       LINODE_REGION: us-sea
       LINODE_CONTROL_PLANE_MACHINE_TYPE: g6-standard-2
       LINODE_MACHINE_TYPE: g6-standard-2
+      WORKER_NODES: ${{ github.head_ref == 'upstream-e2e-tests' && '3' || '0' }}
```
rahulait, line 79: will be changed to main before merge

hack/upstream-e2e-tests.sh
komer3: It might be worth testing since for these test we are not planning on having 3 worker nodes

rahulait: Enabling multiVolume causes some of the tests to fail, they are a bit flaky. I've created a separate ticket to enable these in future once we get them to run successfully everytime with retries. Currently, few tests under it keep on failing with no node available errors and we need to troubleshoot why that happens.

.github/workflows/ci.yml
rahulait: updated

.github/workflows/ci.yml
rahulait: updated

eljohnson92 (approved):

PR #198, opened 2024-08-15
[Fixes] - Addressing test cases we found running sanity tests for Controller Plugin

komer3:
The test cases to address:
- "should fail when requesting to create a volume with already existing name and different capacity"
     - We now throwing error with grpc code AlreadyExist when existing vol capacity differs from requested one.
- "should fail when the volume is already published but is incompatible"
     - Remove the capability for READONLY volumes since we do not support creating read only volumes on the cloud provider side at the moment. If needed, we could provide readonly capability on the Node Plugin side in future.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/internal/driver/capabilities.go
+++ b/internal/driver/capabilities.go
@@ -8,7 +8,6 @@ func ControllerServiceCapabilities() []*csi.ControllerServiceCapability {
 	capabilities := []csi.ControllerServiceCapability_RPC_Type{
 		csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
 		csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME,
-		csi.ControllerServiceCapability_RPC_PUBLISH_READONLY,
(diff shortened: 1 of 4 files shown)


internal/driver/capabilities.go
```diff
@@ -8,7 +8,6 @@ func ControllerServiceCapabilities() []*csi.ControllerServiceCapability {
 	capabilities := []csi.ControllerServiceCapability_RPC_Type{
 		csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
 		csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME,
-		csi.ControllerServiceCapability_RPC_PUBLISH_READONLY,
```
rahulait, line 11: Do we need to remove this capability? Or we are removing this because we don't support read only volumes?

rahulait: Just read the PR description and figured out that we are removing this as we currently don't support it.

rahulait (approved):
LGTM

eljohnson92 (approved):

PR #204, opened 2024-08-20
[Cleanup] Nodepublishvolume() refactor

komer3:
This PR is refactoring one of the Node Plugins grpc methods called NodePublishVolume(). Its purpose is to mount a volume device (block or fs) to a given target directory inside of a particular pod. 

Things that are refactored:
- Moved request obj validation into its own helper func and added unit testing for it
- Moved the FileSystem Interface and its implementation into pkg/mountmanager
    - Extended the interface to accommodate new refactor changes and needs for unit testing 
- Refactored and moved the logic of mounting block volume device into its own func called nodePublishVolumeBlock(). It called if the volume is of type Block.
    - This func implements the logic of creating a file at the target path for bind mounting so pod can directly access the volume device
    - Added unit testing for this func as well  

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


internal/driver/driver_test.go
```diff
@@ -55,7 +55,7 @@ func TestDriverSuite(t *testing.T) {
 	mounter := mountmanager.NewFakeSafeMounter()
 	deviceUtils := mountmanager.NewFakeDeviceUtils()
 	// TODO: Replace by mock implementation later
```
rahulait, line 57:
```suggestion
```

internal/driver/nodeserver_helpers.go
```diff
@@ -114,6 +91,24 @@ func validateNodeExpandVolumeRequest(req *csi.NodeExpandVolumeRequest) error {
...
+	} 
+	if req.GetTargetPath() == "" {
+		return errNoTargetPath
+	}
+	if req.GetVolumeCapability() == nil {
+		return errNoVolumeCapability
+	}
+	return nil
+}
+
```
amold1, line 111: `validateNodeUnpublishVolumeRequest` is a subset of the validation performed here. Can we remove the unpublish function and pass an operation to this function instead?
```suggestion
// validateNodePublishVolumeRequest validates the node publish volume request.
// It checks the volume ID, staging target path, target path, and volume capability in the provided request.
func validateNodePublishVolumeRequest(req *csi.NodePublishVolumeRequest) error {
	if req.GetVolumeId() == "" {
		return errNoVolumeID
	}
	if req.GetStagingTargetPath() == "" {
		return errNoStagingTargetPath
	} 
	if req.GetTargetPath() == "" {
		return errNoTargetPath
	}
	if req.GetVolumeCapability() == nil {
		return errNoVolumeCapability
	}
	return nil
}

```
```
// validateNodePublishVolumeRequest validates the node publish volume request.
// It checks the volume ID and target path for all operations, as well as the staging target path and volume capability ifirn the provided request.
func validateNodePublishVolumeRequest(req *csi.NodePublishVolumeRequest, requestType string) error {
	if req.GetVolumeId() == "" {
		return errNoVolumeID
	}
	if req.GetTargetPath() == "" {
		return errNoTargetPath
	}
	if requestType == "publish" {
		if req.GetStagingTargetPath() == "" {
			return errNoStagingTargetPath
		} 
		if req.GetVolumeCapability() == nil {
			return errNoVolumeCapability
		}
	}
	return nil
}

```

internal/driver/nodeserver.go
```diff
@@ -54,27 +52,22 @@ func (ns *LinodeNodeServer) NodePublishVolume(ctx context.Context, req *csi.Node
 	defer ns.mux.Unlock()
 	klog.V(4).Infof("NodePublishVolume called with req: %#v", req)
 
-	// Validate Arguments
+	// Validate the request object
+	if err := validateNodePublishVolumeRequest(req); err != nil {
+		return nil, err
+	}
+
+	// Path to where we want to mount the volume inside the pod
```
amold1, line 60: Can we assign these variable if and when they are needed in the function? There are multiple return cases before we get to the mounting stage. Just a good practice to avoid allocation memory for stuff that may not be needed

internal/driver/nodeserver_helpers.go
komer3: I can try and combine all of the validation funcs in a following PR to limit the scope

internal/driver/nodeserver.go
komer3: Added the changes

amold1 (approved):

PR #225, opened 2024-09-06
adding automaxproc to auto optimize GOPROC management

prajwalvathreya:
This PR adds the automaxproc library to csi driver

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/main.go
+++ b/main.go
@@ -83,7 +83,6 @@ func main() {
 
 	if maxprocsError != nil {
 		log.Error(maxprocsError, "Failed to set GOMAXPROCS")
-		os.Exit(1)
 	}
 
 	if err := handle(ctx); err != nil {
--- a/main.go
+++ b/main.go
@@ -19,7 +19,6 @@ import (
 	"flag"
 	"fmt"
 	"go.uber.org/automaxprocs/maxprocs"
-	stdlog "log"
 	"os"
 
 	"github.com/ianschenck/envflag"
@@ -77,11 +76,14 @@ func main() {
 	ctx := context.Background()
 	log := logger.NewLogger(ctx)
 	ctx = context.WithValue(ctx, logger.LoggerKey{}, log)
-	undo, maxprocsError := maxprocs.Set(maxprocs.Logger(stdlog.Printf))
-	defer undo()
+	undoMaxprocs, maxprocsError := maxprocs.Set(maxprocs.Logger(func(msg string, keysAndValues ...interface{}) {
+		log.Klogr.WithValues("component", "maxprocs", "version", maxprocs.Version).V(2).Info(fmt.Sprintf(msg, keysAndValues...))
+	}))
+	defer undoMaxprocs()
 
 	if maxprocsError != nil {
-		stdlog.Fatalf("failed to set GOMAXPROCS: %v", maxprocsError)
+		log.Error(maxprocsError, "Failed to set GOMAXPROCS")
+		os.Exit(1)
 	} else if err := handle(ctx); err != nil {
 		log.Error(err, "Fatal error")
 		os.Exit(1)
--- a/main.go
+++ b/main.go
@@ -21,8 +21,8 @@ import (
 	"os"
 
 	"github.com/ianschenck/envflag"
-	"go.uber.org/automaxprocs/maxprocs"
 	"github.com/linode/linodego"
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	"github.com/ianschenck/envflag"
 	"github.com/linode/linodego"
+	_ "go.uber.org/automaxprocs"
```
komer3, line 25: Could you do something like: https://pkg.go.dev/go.uber.org/automaxprocs/maxprocs#example-Logger

Reason: We want to log what maxproc is being set to and also confirm this automaxproc is working

prajwalvathreya: Sure, I can add this and push an update

main.go
```diff
@@ -75,11 +77,16 @@ func main() {
 	ctx := context.Background()
 	log := logger.NewLogger(ctx)
 	ctx = context.WithValue(ctx, logger.LoggerKey{}, log)
+	undo, maxprocsError := maxprocs.Set(maxprocs.Logger(stdlog.Printf))
```
komer3, line 80: Please use the existing logger implementation at pkg/logger

main.go
```diff
@@ -75,11 +77,16 @@ func main() {
...
+	defer undo()
 
-	if err := handle(ctx); err != nil {
+	if maxprocsError != nil {
+		stdlog.Fatalf("failed to set GOMAXPROCS: %v", maxprocsError)
+	} else if err := handle(ctx); err != nil {
```
komer3, line 85: nit: no need for else if here

prajwalvathreya: To handle the maxprocs error this is needed right?

main.go
prajwalvathreya: Sure

main.go
komer3: I meant can just do:

```
if maxprocsError != nil {
  // some code
}

if err := handle(ctx); err != nil {
  // some code
}
```

prajwalvathreya: Cool, I'll update that

komer3 (approved):

main.go
```diff
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"go.uber.org/automaxprocs/maxprocs"
 	"os"
 
 	"github.com/ianschenck/envflag"
```
AshleyDumaine, line 24:
```suggestion
	"os"

	"github.com/ianschenck/envflag"
	"go.uber.org/automaxprocs/maxprocs"
```
imports should be grouped according to https://go.dev/wiki/CodeReviewComments#imports

main.go
```diff
@@ -21,6 +21,7 @@ import (
...
+	"go.uber.org/automaxprocs/maxprocs"
 	"github.com/linode/linodego"
```
AshleyDumaine, line 25:
```suggestion
	"github.com/linode/linodego"
	"go.uber.org/automaxprocs/maxprocs"
```
sorry, one more tweak since the groups should be in alphabetical order.

prajwalvathreya: Sure!

main.go
```diff
@@ -75,11 +76,21 @@ func main() {
...
 	log := logger.NewLogger(ctx)
 	ctx = context.WithValue(ctx, logger.LoggerKey{}, log)
+	undoMaxprocs, maxprocsError := maxprocs.Set(maxprocs.Logger(func(msg string, keysAndValues ...interface{}) {
+		log.Klogr.WithValues("component", "maxprocs", "version", maxprocs.Version).V(2).Info(fmt.Sprintf(msg, keysAndValues...))
+	}))
+	defer undoMaxprocs()
+
+	if maxprocsError != nil {
+		log.Error(maxprocsError, "Failed to set GOMAXPROCS")
+		os.Exit(1)
```
AshleyDumaine, line 86: I don't think we should necessarily exit if we can't set `GOMAXPROCS` since that could result in a CrashLoopBackoff where there was none prior without this, but we should keep the error log.

prajwalvathreya: I agree. This would introduce a new problem. I'll fix that and push the changes.

komer3 (approved):

eljohnson92 (approved):

Issue #224, opened 2024-09-06
Encrypted PV gets corrupted when statefulSet is restarted multiple times.


nishantbansal7869:
Linode CSI driver version: 0.8.3
When a statefulSet with encrypted PV is restarted multiple times, then the PV get corrupted. It gets attached to the new node, but while formatting and mounting the PV, it throws the following error:

024-09-06T13:43:58+05:30 Writing superblocks and filesystem accounting information:  0/80mkfs.ext4: I/O error while writing out and closing file system
2024-09-06T13:43:58+05:30 ) 
2024-09-06T13:43:58+05:30 E0906 08:13:58.812779       1 utils.go:30] GRPC error: rpc error: code = Internal desc = Failed to format and mount device from ("/dev/disk/by-id/scsi-0Linode_Volume_pvca998862010864e82") to ("/var/lib/kubelet/plugins/kubernetes.io/csi/linodebs.csi.linode.com/946abaed08141c01fc687279fdf7a52f48c0124b03163bc70b89856f2732f001/globalmount") with fstype ("ext4") and options ([]): format of disk "/dev/mapper/pvca998862010864e82" failed: type:("ext4") target:("/var/lib/kubelet/plugins/kubernetes.io/csi/linodebs.csi.linode.com/946abaed08141c01fc687279fdf7a52f48c0124b03163bc70b89856f2732f001/globalmount") options:("defaults") errcode:(exit status 1) output:(mke2fs 1.47.0 (5-Feb-2023)
2024-09-06T13:43:58+05:30 Warning: could not erase sector 2: I/O error
2024-09-06T13:43:58+05:30 Creating filesystem with 2617344 4k blocks and 655360 inodes
2024-09-06T13:43:58+05:30 Filesystem UUID: 2ffe1040-4c33-49d6-b35d-1e9f00b76f68
2024-09-06T13:43:58+05:30 Superblock backups stored on blocks: 
2024-09-06T13:43:58+05:30 	32768, 98304, 163840, 229376, 294912, 819200, 884736, 1605632
2024-09-06T13:43:58+05:30 
2024-09-06T13:43:58+05:30 Allocating group tables:  0/80     done                            
2024-09-06T13:43:58+05:30 Warning: could not read block 0: I/O error
2024-09-06T13:43:58+05:30 Warning: could not erase sector 0: I/O error
2024-09-06T13:43:58+05:30 Writing inode tables:  0/80     done                            
[...]

PR #228, opened 2024-09-10
[Testing] Add New E2E Test Case & Enhance E2E Testing Configuration

komer3:
## Reason for this PR:
New E2E test (pod-pvc-luks-mov-volume) being added to make sure we check for this edge case: https://github.com/linode/linode-blockstorage-csi-driver/issues/224

- Added E2E_SELECTOR environment variable to CI workflow for dynamic test selection based on branch.
- Updated e2e-test command to utilize E2E_SELECTOR for improved test execution control.
- Added labels ('all' and 'quick') to multiple test definitions for better categorization and selection.
- Renamed 'pod-pvc-expand-storage' test to 'pod-pvc-expand-storage-size' for clarity.

## How to run the new test and verify:
1. Create a 1 control plane and 1 worker node cluster with CSI driver deployed
    `CONTROLPLANE_NODES=1 WORKER_NODES=1 devbox run capl-cluster`
2. Make sure you are running CSI driver image with tag `statefulset-encrypt-pv-fix`. You can edit the resource then rollout restart. Link to image: https://hub.docker.com/layers/linode/linode-blockstorage-csi-driver/statefulset-encrypt-pv-fix/images/sha256-0da0820bbaf72a8becd5ba4cd998ba8f04e95a682415fdd8cee9a9dfa504dbf8?context=explore
3. Run the e2e-test command:
    `E2E_SELECTOR=volumemove CONTROLPLANE_NODES=1 WORKER_NODES=1 devbox run e2e-test`

For more info on e2e test setup, please refer: https://github.com/linode/linode-blockstorage-csi-driver/tree/main/tests/e2e

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


tests/e2e/test/pod-pvc-luks-mov-volume/chainsaw-test.yaml
amold1, line 1: This LGTM - Can we add a step to check if the original mapping on CP node is removed after the pod has moved to the worker node? That way we can be sure that when the pod is moved back to the CP from worker it is not using any old mapper

komer3: We could but it would take more effort than it might be worth. Being able to mount back to control plane node in the 3rd step confirms that there was no lingering mapping (current branch will actually fail since it doesn't have the fix, I'll enable the test to show it). When we take a look at the CSI driver Node Server logs, we are able to confirm that new mapper was created and can be mounted successfully. 

Lmk what you think!

amold1: SGTM

amold1 (approved):

eljohnson92 (approved):

PR #232, opened 2024-09-12
replace using cryptsetup binary for LUKS encryption with cryptsetup go pkg

amold1:
### Summary:

This change replaces the use of cryptsetup binary for LUKS encryption with [cryptsetup go pkg](https://pkg.go.dev/github.com/martinjungblut/go-cryptsetup).
The rationale for doing this is eloquently stated in the go pkg docs [here](https://pkg.go.dev/github.com/martinjungblut/go-cryptsetup#section-readme:~:text=Deactivating%20devices-,Rationale,-A%20number%20of)

### Related Changes;
Given this pkg uses underlying C libraries that are available only for linux systems and not macos, we are now moving our make targets to use docker containers(with an image created from dev dockerfile)


internal/driver/luks.go
komer3, line 1: Could we add unit testing for the new luks functions?

internal/driver/luks.go
```diff
@@ -115,179 +169,70 @@ func getLuksContext(secrets map[string]string, context map[string]string, lifecy
...
-
-func (e *Encryption) luksOpen(ctx LuksContext, volume string) error {
-	// check if the luks volume is already open
-	if _, err := e.FileSystem.Stat("/dev/mapper/" + ctx.VolumeName); !e.FileSystem.IsNotExist(err) {
-		klog.V(4).Infof("luks volume is already open %s", volume)
-		return nil
-	}
-
-	cryptsetupCmd, err := e.getCryptsetupCmd()
+	klog.V(4).Info("Formatting luks device ", newLuksDevice.device)
```
komer3, line 190: Could you replace the klog with new logger implementation in `pkg/logger`

internal/driver/nodeserver.go
```diff
@@ -140,24 +140,25 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 	ns.mux.Lock()
 	defer ns.mux.Unlock()
 
 	// Validate request object
 	log.V(4).Info("Validating request", "volumeID", volumeID)
-	if err := validateNodeUnpublishVolumeRequest(ctx, req); err != nil {
+	err := validateNodeUnpublishVolumeRequest(ctx, req)
+	if err != nil {
 		return nil, err
 	}
```
komer3, line 155: nit: Can we use this pattern if a func is returning back just an err?

`if err := validateNodeUnpublishVolumeRequest(ctx, req); err != nil {
      return nil, err
}`

internal/driver/nodeserver.go
```diff
@@ -140,24 +140,25 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
 	log, ctx, done := logger.GetLogger(ctx).WithMethod("NodeUnpublishVolume")
 	defer done()
-	
+
 	volumeID := req.GetVolumeId()
-	targetPath := req.GetTargetPath()
-	log.V(2).Info("Processing request", "volumeID", volumeID, "targetPath", targetPath)
-	
+	log.V(2).Info("Processing request", "volumeID", volumeID)
```
komer3, line 145: question: Any reason for removing targetPath here?

internal/driver/nodeserver.go
```diff
@@ -241,28 +242,28 @@ func (ns *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 	defer done()
 
 	volumeID := req.GetVolumeId()
-	stagingTargetPath := req.GetStagingTargetPath()
-	log.V(2).Info("Processing request", "volumeID", volumeID, "stagingTargetPath", stagingTargetPath)
+	log.V(2).Info("Processing request", "volumeID", volumeID)
```
komer3, line 245: question: Any reason for removing stagingTargetPath here?

internal/driver/nodeserver.go
```diff
@@ -241,28 +242,28 @@ func (ns *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
...
+	if err != nil {
+		return nil, errInternal("NodeUnstageVolume failed to unmount at path %s: %v", req.GetStagingTargetPath(), err)
 	}
 
 	// If LUKS volume is used, close the LUKS device
-	log.V(4).Info("Closing LUKS device", "volumeID", volumeID, "stagingTargetPath", stagingTargetPath)
-	if err := ns.closeLuksMountSource(ctx, volumeID); err != nil {
-		return nil, errInternal("Failed to close the luks volume %s: %v", volumeID, err)
+	log.V(4).Info("Closing LUKS device", "volumeID", volumeID, "stagingTargetPath", req.GetStagingTargetPath())
+	if err := ns.closeLuksMountSource(ctx, req.GetVolumeId()); err != nil {
```
komer3, line 265: nit: we can use volumeID var here instead of `req.GetVolumeId()`

internal/driver/nodeserver_helpers.go
```diff
@@ -328,39 +328,24 @@ func (ns *NodeServer) mountVolume(ctx context.Context, devicePath string, req *c
 // If not, it formats the device using the provided LuksContext.
 // Finally, it prepares the LUKS volume for mounting.
 func (ns *NodeServer) prepareLUKSVolume(ctx context.Context, devicePath string, luksContext LuksContext) (string, error) {
+	var luksSource string
+	var err error
```
komer3, line 332: Question: any reason why we are defining vars here instead of using `:=` below?

internal/driver/nodeserver_helpers.go
```diff
@@ -328,39 +328,24 @@ func (ns *NodeServer) mountVolume(ctx context.Context, devicePath string, req *c
...
 
 	// LUKS encryption enabled, check if the volume needs to be formatted.
 	log.V(4).Info("LUKS encryption enabled")
 
-	// Validate if the device is formatted with LUKS encryption or if it needs formatting.
-	formatted, err := ns.encrypt.blkidValid(devicePath)
-	if err != nil {
-		return "", errInternal("Failed to validate blkid (%q): %v", devicePath, err)
-	}
-	log.V(4).Info("Disk format check", "devicePath", devicePath, "formatted", formatted)
```
komer3, line 342: question: with new luks implementation, do we not need to check if the device is formatted already?

internal/driver/nodeserver_helpers.go
```diff
@@ -372,33 +357,19 @@ func (ns *NodeServer) prepareLUKSVolume(ctx context.Context, devicePath string,
 // Returns an error if any operation fails during the process.
 func (ns *NodeServer) closeLuksMountSource(ctx context.Context, volumeID string) error {
 	log := logger.GetLogger(ctx)
-	log.V(4).Info("Entering closeLuksMountSource", "volumeID", volumeID)
 
-	mountSource, err := ns.getMountSource(ctx, volumeID)
+	volumeName, err := ns.getMountSource(ctx, volumeID)
 	if err != nil {
-		return errInternal("closeLuksMountSource failed to get mount source %s: %v", volumeID, err)
+		return fmt.Errorf("Could not get volumename in pvcxxxxxxxxx format from given volumeID: %s", volumeID)
```
komer3, line 363: This error message is not necessarily true. We just split the volumeID string into 2 parts but delimiter `-`. We aren't checking if the second string has `pvc` as a prefix. 

Reason: CSI-sanity tests send requested to CSI driver that does not conform with `pvc` prefix format

internal/driver/nodeserver_helpers.go
```diff
@@ -328,39 +328,24 @@ func (ns *NodeServer) mountVolume(ctx context.Context, devicePath string, req *c
 // If not, it formats the device using the provided LuksContext.
 // Finally, it prepares the LUKS volume for mounting.
 func (ns *NodeServer) prepareLUKSVolume(ctx context.Context, devicePath string, luksContext LuksContext) (string, error) {
```
komer3, line 330: nit: Could we change the name of this func to something more focused? example: formatLuksVolume()

internal/driver/luks.go
```diff
@@ -22,16 +22,68 @@ limitations under the License.
...
+	if err != nil {
+		return LuksDevice{}, err
+	}
+	return LuksDevice{device: dev}, nil
+}
+
+func NewCryptSetup() CryptSetup {
+	return CryptSetup{}
+}
+
```
komer3, line 86: question: What do you think about moving this closer to the interface in `pkg/cryptsetup-client`?

internal/driver/luks.go
amold1: The new implementation requires the context if I am not mistaken. Do you think we should add the context as an argument to `luksFormat` and `luksClose`?

internal/driver/nodeserver.go
amold1: agree. done

internal/driver/nodeserver.go
amold1: added it back

internal/driver/nodeserver.go
amold1: added it back

internal/driver/nodeserver.go
amold1: done

internal/driver/nodeserver_helpers.go
amold1: fixed

internal/driver/nodeserver_helpers.go
amold1: added a check now

internal/driver/nodeserver_helpers.go
amold1: reverted

internal/driver/nodeserver_helpers.go
amold1: implemented

internal/driver/luks.go
amold1: good call... moved

internal/driver/luks.go
amold1: we have in the new test module

internal/driver/luks.go
komer3: Yup, we should add the context and get the logger from it. This way we are able to sort of stack trace in the logs whenever a grpc func is called!

eljohnson92 (approved):

Dockerfile
```diff
@@ -10,17 +10,18 @@ COPY go.sum .
...
 
 FROM alpine:3.18.4
 LABEL maintainers="Linode"
 LABEL description="Linode CSI Driver"
 
 COPY --from=builder /bin/linode-blockstorage-csi-driver /linode
 
-RUN apk add --no-cache e2fsprogs findmnt blkid cryptsetup xfsprogs
-
+RUN apk add --no-cache e2fsprogs findmnt blkid cryptsetup xfsprogs lsblk
```
komer3, line 25: Do we still need this lsblk cryptsetup?

internal/driver/nodeserver.go
```diff
@@ -240,29 +240,30 @@ func (ns *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 	log, ctx, done := logger.GetLogger(ctx).WithMethod("NodeUnstageVolume")
 	defer done()
 
-	volumeID := req.GetVolumeId()
 	stagingTargetPath := req.GetStagingTargetPath()
-	log.V(2).Info("Processing request", "volumeID", volumeID, "stagingTargetPath", stagingTargetPath)
+	volumeID := req.GetVolumeId()
+	log.V(2).Info("Processing request", "volumeID", volumeID)
```
komer3, line 245: nit: could we add the stagingTargetPath back into the log line

internal/driver/nodeserver_helpers.go
```diff
@@ -314,56 +316,55 @@ func (ns *NodeServer) mountVolume(ctx context.Context, devicePath string, req *c
...
 	}
-	log.V(4).Info("Disk format check", "devicePath", devicePath, "formatted", formatted)
 
-	// If the device is not, format it.
-	if !formatted {
-		log.V(4).Info("luks volume now formatting: ", "devicePath", devicePath)
+	// Validate the LUKS context.
+	if err = luksContext.validate(); err != nil {
+		return "", errInternal("Failed to luks format validation (%q): %v", devicePath, err)
+	}
```
komer3, line 350: Any reason why we are validating the luksContext twice?

internal/driver/nodeserver_helpers.go
```diff
@@ -314,56 +316,55 @@ func (ns *NodeServer) mountVolume(ctx context.Context, devicePath string, req *c
...
-	// LUKS encryption enabled, check if the volume needs to be formatted.
-	log.V(4).Info("LUKS encryption enabled")
+	// Validate the LUKS context.
+	if err = luksContext.validate(); err != nil {
+		return "", errInternal("Failed to luks format validation (%q): %v", devicePath, err)
+	}
 
-	// Validate if the device is formatted with LUKS encryption or if it needs formatting.
+	// LUKS encryption enabled, check if the volume needs to be formatted.
 	formatted, err := ns.encrypt.blkidValid(devicePath)
```
komer3, line 342: nit: if possible could you add `blkidvalid()` to logging by passing the ctx in and updating it?

komer3 (commented):
Just got some small nit comments but otherwise logically things look good!

Great work!!

Dockerfile
amold1: nope... removed it

internal/driver/nodeserver_helpers.go
amold1: done

internal/driver/nodeserver_helpers.go
amold1: nope... removed the initial check as we dont need it if blkid returns error

komer3 (dismissed):
LGTM!

komer3 (approved):

eljohnson92 (approved):

PR #265, opened 2024-09-26
[Update] Bumping up the version for side cars to latest available versions

komer3:
In this PR we are removing args for the CSI driver image, the reason for that is we have moved away from setting flags for config and are supporting setting that using env var. 

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/deploy/kubernetes/base/ds-csi-linode-node.yaml
+++ b/deploy/kubernetes/base/ds-csi-linode-node.yaml
@@ -35,7 +35,7 @@ spec:
       hostNetwork: true
       containers:
         - name: csi-node-driver-registrar
-          image: registry.k8s.io/sig-storage/csi-node-driver-registrar:v1.3.0
+          image: registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.12.0
           args:
             - "--v=2"
             - "--csi-address=$(ADDRESS)"
--- a/deploy/kubernetes/base/ds-csi-linode-node.yaml
+++ b/deploy/kubernetes/base/ds-csi-linode-node.yaml
@@ -54,7 +54,7 @@ spec:
               mountPath: /csi
             - name: registration-dir
               mountPath: /registration
-        - name: csi-linode-plugin
+        - name: linode-csi-plugin
           image: linode/linode-blockstorage-csi-driver:latest
           args :
             - "--v=2"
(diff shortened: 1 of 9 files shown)


amold1 (approved):

internal/driver/deploy/releases/linode-blockstorage-csi-driver.yaml
```diff
@@ -471,7 +462,7 @@ spec:
             secretKeyRef:
               key: token
               name: linode
-        image: linode/linode-blockstorage-csi-driver:latest
+        image: linode/linode-blockstorage-csi-driver:version-upgrade
```
AshleyDumaine, line 465: Should change this back to latest right?

komer3: good catch!

helm-chart/csi-driver/values.yaml
```diff
@@ -31,17 +31,17 @@ defaultStorageClass: linode-block-storage-retain
...
 csiAttacher:
   image: registry.k8s.io/sig-storage/csi-attacher
-  tag: v3.3.0
+  tag: v4.7.0
   pullPolicy: IfNotPresent
 
 csiResizer:
   image: registry.k8s.io/sig-storage/csi-resizer
-  tag: v1.3.0
+  tag: v1.12.0
```
rahulait, line 44: just double checking, should this be same as that of csi-node-driver-registrar which is v2.12.0 or it is v1.12.0 only?

rahulait: Seems to be correct, latest for csi-resizer is v1.12.0 only and for csi-node-driver-registrar its v2.12.0

rahulait (dismissed):
lgtm

amold1 (approved):

deploy/kubernetes/base/ds-csi-linode-node.yaml
```diff
@@ -35,7 +35,7 @@ spec:
       hostNetwork: true
       containers:
         - name: csi-node-driver-registrar
-          image: registry.k8s.io/sig-storage/csi-node-driver-registrar:v1.3.0
+          image: registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.12.0
           args:
             - "--v=2"
             - "--csi-address=$(ADDRESS)"
```
rahulait, line 41: We should also change `csi-linode-node` to `linode-csi-node` in this file in case we want to keep things consistent.

rahulait: It looks like we'll need to change bunch of other files as well if we want to switch to `linode-*` naming scheme

amold1 (approved):

rahulait (approved):

amold1 (approved):

PR #278, opened 2024-10-02
[Feat] Node Server Metrics Implementation 

prajwalvathreya:
This PR is to get the metrics of the functions performed by the node server. The functions for which the metrics are implemented are:

NodePublishVolume
NodeUnpublishVolume
NodeStageVolume
NodeUnstageVolume
NodeExpandVolume

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


internal/driver/server.go
```diff
@@ -45,13 +50,15 @@ func NewNonBlockingGRPCServer() NonBlockingGRPCServer {
...
-	server *grpc.Server
+	wg            sync.WaitGroup
+	server        *grpc.Server
+	metricsServer *http.Server
 }
 
 func (s *nonBlockingGRPCServer) Start(endpoint string, ids csi.IdentityServer, cs csi.ControllerServer, ns csi.NodeServer) {
 	s.wg.Add(1)
 	go s.serve(endpoint, ids, cs, ns)
+	go s.startMetricsServer(":8081")
```
komer3, line 61: Can we make this dynamic and get the value from the ENV var? May be also only enable metric server if a flag is set to true

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	SuccessTrue  = "true"
+	SuccessFalse = "false"
+)
+
+var (
```
komer3, line 12: Could you add some comments about these prom metrics? Would be helpful in future when trying to understand this

internal/driver/nodeserver.go
```diff
@@ -86,15 +88,22 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 
+	success := metrics.SuccessTrue
+
 	// Validate the request object
 	log.V(4).Info("Validating request", "volumeID", volumeID)
 	if err := validateNodePublishVolumeRequest(ctx, req); err != nil {
+		success = metrics.SuccessFalse
+		// Record failure metric before returning
+		metrics.NodePublishTotal.WithLabelValues(success).Inc()
+		metrics.NodePublishDuration.WithLabelValues(success).Observe(time.Since(start).Seconds())
```
komer3, line 106: Looks like these lines are being used multiple times, we should consider encapsulating it in some func or var?

pkg/metrics/metrics.go
prajwalvathreya: Sure

internal/driver/nodeserver.go
prajwalvathreya: Yes, this makes sense. I have pushed changes to encapsulate this.

internal/driver/server.go
amold1: +1 for this

internal/driver/nodeserver.go
```diff
@@ -86,15 +88,20 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 	defer done()
 
+	start := time.Now()
 	volumeID := req.GetVolumeId()
 	log.V(2).Info("Processing request", "volumeID", volumeID)
 
 	ns.mux.Lock()
 	defer ns.mux.Unlock()
 
+	success := metrics.SuccessTrue
```
amold1, line 98: what is this `success` variable indicating? That the controller was successful in publishing the volume(basically whatever task it was attempting)?

prajwalvathreya: Yes, the variable indicates if the function was completed. `true` is completion with no errors. `false` indicates that there was an error and it's trying to execute the function again until completion.

internal/driver/server.go
prajwalvathreya: I've pushed the changes to accept flags to enable the metrics

internal/driver/nodeserver.go
amold1: The function completes on line 156 though. Or Am I missing something?
Also can we rename `start` and `success` in all the methods to something more descriptive?

prajwalvathreya: Yes, it does. So initially the success variable is set to `true`, but on each failure it's set as `false` and stored into prometheus's registrar.  If the execution goes to line 156 without any errors, success will be `true` meaning the function was completed.

Sure,  I'll rename the variables.

helm-chart/csi-driver/templates/csi-linode-controller-metrics.yaml
komer3, line 1: Are we going to create a service for csi driver metrics?

helm-chart/csi-driver/templates/csi-linode-controller.yaml
komer3, line 1: Don't we need to add the new metrics args here? Otherwise we wont be able to enable metrics for our CSI driver?

helm-chart/csi-driver/templates/csi-linode-controller-metrics.yaml
```diff
@@ -58,8 +58,29 @@ spec:
...
+    prometheus.io/path: '/metrics'
+spec:
+  selector:
+    app: csi-linode-node
+  ports:
+    - name: node-server-metrics-port
+      port: {{ .Values.metricsPort }}
+      targetPort: {{ .Values.metricsPort }}
+      protocol: TCP
+{{- end }}
```
komer3, line 86: nit: please add new line at the end of file

helm-chart/csi-driver/templates/daemonset.yaml
```diff
@@ -55,7 +55,7 @@ spec:
         - name: ENABLE_METRICS
           value: {{ .Values.enableMetrics | quote}}
         - name: METRICS_ADDRESS
-          value: {{ .Values.metricsAddress | quote}}
+          value: {{ .Values.metricsPort | quote}}
```
komer3, line 58: question: any reason to use `quote` here?

internal/driver/nodeserver.go
```diff
@@ -86,15 +88,20 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 	defer done()
 
+	functionStartTime := time.Now()
 	volumeID := req.GetVolumeId()
 	log.V(2).Info("Processing request", "volumeID", volumeID)
 
 	ns.mux.Lock()
 	defer ns.mux.Unlock()
 
+	functionStatus := metrics.Completed
```
komer3, line 98: Do we need to have this `functionStatus` var at all? Why don't we just pass the value directly to `RecordMetrics()`?

prajwalvathreya: Yeah! That should work!

komer3 (approved):
LGTM! Great work on this!

PR #280, opened 2024-10-04
[Feat] Multi region support (Topology Aware Provisioning)

komer3:
## Overview

This PR enhances the CSI driver's controller server to provision volumes in regions outside of its deployment region. The improvement allows for more flexible and efficient volume management across different geographical locations.

## Key Concepts

- **Remote Region**: A region where worker nodes are deployed, separate from the control plane region. 
  Example: Control plane in us-ord, worker node in us-central.

## Current Deployment Setup

- Controller server: Deployed on worker nodes
- Node servers: Deployed as a DaemonSet on all nodes

### 1. Dynamic Region Detection

- **Previously**: Used metadata service in the controller service class to determine the region.
- **Now**: Utilizes topology-based dynamic provisioning.

### 2. Volume Provisioning Process

- External-provisioner makes an API call to the api-server to retrieve the Linode region.
- This region information is included in the CreateVolumeRequest object during the CreateVolume gRPC call.
- Enables volume provisioning in the region where the pod is scheduled.

### 3. Volume Binding Mode

- Changed from 'Immediate' to 'WaitForFirstConsumer'.
- Allows the provisioner time to gather necessary region data before volume creation.

## Benefits

- Improved flexibility in volume provisioning across different regions.
- Better alignment of volume location with pod scheduling.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


docs/topology-aware-provisioning.md
```diff
@@ -0,0 +1,80 @@
...
+  - name: csi-volume
+    persistentVolumeClaim:
+      claimName: pvc-filesystem
+```
+
+This example demonstrates how to set up topology-aware provisioning using the Linode Block Storage CSI Driver. The StorageClass defines the provisioner and reclaim policy, while the PersistentVolumeClaim requests storage from this class. The Pod specification shows how to use the PVC and includes a node selector for region-specific deployment.
+
+#### Provisioning Process
+
+1. CO determines required topology based on application needs and cluster layout.
```
tchinmai7, line 74: can you explain what CO is?

internal/driver/controllerserver_helper.go
```diff
@@ -167,19 +167,27 @@ func (cs *ControllerServer) getContentSourceVolume(ctx context.Context, contentS
...
-	if volumeData.Region != cs.metadata.Region {
+	// Check if the source volume's region matches the server's metadata region
+	// If no topology is specified, the source volume must be in the same region as the server's metadata region
+	if accessibilityRequirements == nil && volumeData.Region != cs.metadata.Region {
 		return nil, errRegionMismatch(volumeData.Region, cs.metadata.Region)
 	}
 
+	// If a topology is specified, the source volume must be in the same region as the specified topology
+	if accessibilityRequirements != nil {
+		if volumeData.Region != accessibilityRequirements.GetPreferred()[0].GetSegments()[VolumeTopologyRegion] {
```
tchinmai7, line 178: is this guaranteed to be an array of len > 0  && is the key guaranteed to exist?

komer3: Its pretty much guaranteed to exist and the first key is going to be the region we want to use.

for more info check this link out! https://kubernetes-csi.github.io/docs/topology.html?highlight=VOLUME_ACCESSIBILITY_CONSTRAINTS#implementing-topology-in-your-csi-driver

tchinmai7: I was curious because `getRegionFromTopology` operates on the same accessibilityRequirements but includes all those checks, Thanks for the link - will check it out!

docs/topology-aware-provisioning.md
komer3: Updated the doc. Hope its more clear now! Let me know if there is anything else that is unclear :)

docs/topology-aware-provisioning.md
```diff
@@ -0,0 +1,79 @@
...
+    - mountPath: /data
+      name: csi-volume
+  volumes:
+  - name: csi-volume
+    persistentVolumeClaim:
+      claimName: pvc-filesystem
+```
+
+This example demonstrates how to set up topology-aware provisioning using the Linode Block Storage CSI Driver. The StorageClass defines the provisioner and reclaim policy, while the PersistentVolumeClaim requests storage from this class. The Pod specification shows how to use the PVC and includes a node selector for region-specific deployment.
+
```
tchinmai7, line 71: should we also mention that the cluster itself must be started with `--feature-gates=CSINodeInfo=true ` ? https://kubernetes-csi.github.io/docs/topology.html?highlight=VOLUME_ACCESSIBILITY_CONSTRAINTS#kubernetes-cluster-setup

internal/driver/controllerserver_helper.go
tchinmai7: could also use that here - to validate

komer3: The only time that is not passed (accessibilityRequirements) is when you have a broken multi region cluster (I had this issue while trying to make konnectivity work). In that senario, I'm falling back to using the region from the metadata (region where controller server is running)

docs/topology-aware-provisioning.md
komer3: Yeah good point. Let me add that as well. 

Also, that feature flag is going to be turn on by default now with future csi releases

internal/driver/controllerserver_helper.go
komer3: Good catch. Let me include those checks here as well

komer3: Updated it!

docs/topology-aware-provisioning.md
```diff
@@ -0,0 +1,88 @@
...
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  storageClassName: linode-block-storage-topology-aware
+```
+
+> **Important**: The `volumeBindingMode: WaitForFirstConsumer` setting is crucial for topology-aware provisioning. It delays volume binding and creation until a pod using the PVC is created. This allows the system to consider the pod's scheduling requirements and node assignment when selecting the most appropriate storage location, ensuring optimal data locality and performance.
```
srust, line 39: What if `WaitForFirstConsumer` is not specified. Presumably the pod would more often than not simply not start because the volume is in a different region. Is there a "default" region for the `Immediate` mode?

komer3: Yes there is! The default/fallback region is the region where the csi controller server is deployed (currently in the worker nodes region).

eljohnson92 (approved):

PR #299, opened 2024-10-30
[feat] Enable support for encrypted volumes for CSI driver 

prajwalvathreya:
### General:
This PR focuses on enabling support for encrypted volumes for the CSI driver. 

It addresses the following tasks:

- creating an encrypted volume (region specific) 
- attaching the volume to a node (region specific) 
- handling errors in case of mismatch in attaching a volume to a node

### Pull Request Guidelines:

1. [ ] Have you added tests? 


helm-chart/csi-driver/templates/csi-linode-controller.yaml
```diff
@@ -108,6 +108,8 @@ spec:
               value: {{ .Values.enableMetrics | quote}}
             - name: METRICS_PORT
               value: {{ .Values.metricsPort | quote}}
+            - name: VOLUME_ENCRYPTION
```
eljohnson92, line 111: is there a reason we need this set as an env variable if it can be enabled per storage class?

helm-chart/csi-driver/values.yaml
```diff
@@ -11,6 +11,9 @@ enableMetrics: false
 # default metrics address port
 metricsPort: 8081
 
+# enable encryption for volumes
```
eljohnson92, line 14:
```suggestion
# enable encryption for volumes by default
```
I think this means it is enabled by default?

prajwalvathreya: Since encryption is disabled by default I meant to write "Enable this variable for encryption on volumes". I could just comment "Set this variable to true for encryption" as it would be more descriptive?

helm-chart/csi-driver/templates/csi-linode-controller.yaml
prajwalvathreya: I am setting this env variable to use it for decision-making in controllerserver.go and controllerserver_helper.go . Basically this variable helps in the createVolume and attachVolume steps.

eljohnson92: we should be using `linodebs.csi.linode.com/encrypted: "true"` to decide if a volume is being encrypted instead of this environment variable

prajwalvathreya: Sure, I've updated the implementation to use this

internal/driver/controllerserver_helper.go
```diff
@@ -272,6 +291,36 @@ func (cs *ControllerServer) createLinodeVolume(ctx context.Context, label string
...
+	log.V(4).Info("Checking if encryption is supported for region", "region", region)
+
+	// Fetch the list of regions from the Linode API
+	regions, err := cs.client.ListRegions(ctx, nil)
+	if err != nil {
+		return false, errInternal("failed to fetch regions: %v", err)
+	}
+
+	// Look for the specified region and check if encryption is supported.
+	for i := range regions {
```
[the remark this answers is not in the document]
prajwalvathreya: Sure, I will make this change

helm-chart/csi-driver/templates/linode-block-storage-retain.yaml
```diff
@@ -7,10 +7,13 @@ metadata:
...
     storageclass.kubernetes.io/is-default-class: "true"
 {{- end }}
-{{- if .Values.volumeTags }}
 parameters:
+{{- if .Values.volumeTags }}
   linodebs.csi.linode.com/volumeTags: {{ join "," .Values.volumeTags }}
 {{- end}}
+{{- if .Values.volumeEncryption }}
+  linodebs.csi.linode.com/encrypted: "true"
+{{- end}}
```
komer3, line 16: We don't want to render parameters if both volume tags and volumeEncryption is not passed in which is highly likely since they are both optional

helm-chart/csi-driver/templates/linode-block-storage.yaml
```diff
@@ -7,9 +7,12 @@ metadata:
...
     storageclass.kubernetes.io/is-default-class: "true"
 {{- end }}
-{{- if .Values.volumeTags }}
 parameters:
+{{- if .Values.volumeTags }}
   linodebs.csi.linode.com/volumeTags: {{ join "," .Values.volumeTags }}
 {{- end}}
+{{- if .Values.volumeEncryption }}
+  linodebs.csi.linode.com/encrypted: "true"
+{{- end}}
```
komer3, line 16: Same here

helm-chart/csi-driver/values.yaml
```diff
@@ -11,6 +11,9 @@ enableMetrics: false
 # default metrics address port
 metricsPort: 8081
 
+# enable encryption for volumes by default
+volumeEncryption: false
+
```
komer3, line 16: This comment is conflicting.

Question: Why are we setting this value at all? SInce its optional why not just keep it commented?

internal/driver/controllerserver.go
```diff
@@ -93,7 +93,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 
 	// Create the volume
-	vol, err := cs.createAndWaitForVolume(ctx, volName, sizeGB, req.GetParameters()[VolumeTags], sourceVolInfo, accessibilityRequirements)
+	vol, err := cs.createAndWaitForVolume(ctx, volName, req.GetParameters()[VolumeTags], req.GetParameters()[VolumeEncryption], sizeGB, sourceVolInfo, accessibilityRequirements)
```
komer3, line 96: Nit: we can just pass req.GetParameters() here and handle getting the volumetags and volumEncryption in the func. That would be much cleaner approach

internal/driver/controllerserver_helper.go
```diff
@@ -250,11 +253,27 @@ func (cs *ControllerServer) createLinodeVolume(ctx context.Context, label string
...
+		if err != nil {
+			return nil, err
+		}
+		if !supported {
+			return nil, errInternal("Volume encryption is not supported in the %s region", region)
+		}
+		encryptionStatus = "enabled"
+	} else {
+		encryptionStatus = "disabled"
+	}
```
komer3, line 269: I think it makes more sense to have this logic inside prepareVolumeParams() helper since we are essentially extracting info from params, validating, and then setting encryption to true.

internal/driver/controllerserver_helper.go
```diff
@@ -272,6 +291,30 @@ func (cs *ControllerServer) createLinodeVolume(ctx context.Context, label string
...
+
+	// Get the specifications of specified region from Linode API
+	regionDetails, err := cs.client.GetRegion(ctx, region)
+	if err != nil {
+		return false, errInternal("failed to fetch region %s: %v", region, err)
+	}
+
+	// Check if encryption is supported in the specified region
+	for _, capability := range regionDetails.Capabilities {
+		if capability == "Block Storage Encryption" {
```
komer3, line 308: I'm a bit concerned about this string comparison here. We could run into some issues down the line. Might be worth it to use regex to look for encryption enablement?

internal/driver/controllerserver_helper.go
```diff
@@ -649,8 +692,31 @@ func (cs *ControllerServer) attachVolume(ctx context.Context, volumeID, linodeID
...
+		linode, err = cs.client.GetInstance(ctx, linodeID)
+		if err != nil {
+			return status.Errorf(codes.Internal, "failed to retrieve Linode instance details: %v", err)
+		}
+
+		// Ensure the encrypted volume and Linode are in the same region
+		if volume.Region != linode.Region {
+			return status.Errorf(codes.FailedPrecondition, "encrypted volume and Linode instance must be in the same region for attachment")
+		}
+	}
```
komer3, line 716: This should be done inside getAndValidateVolume() helper. That way you don't have to make these 2 API calls here. We can reduce redundancy here.

We have the linode instance obj and volume obj available in getAndValidateVolume() plus logically makes sense to put this validation there. 

attachVolume() should only be for attaching logic not validation

internal/driver/controllerserver.go
komer3: Could you also set the volumeContext to include the encryption is set to true? This will be useful to double check in the controllerpublishvolume() grpc func!

helm-chart/csi-driver/templates/linode-block-storage-retain.yaml
prajwalvathreya: Sure, I'll change that

internal/driver/controllerserver.go
prajwalvathreya: Yup, working on these changes.

internal/driver/controllerserver_helper.go
prajwalvathreya: So I tried moving this into prepareVolumeParams() but it's a contradicting change with the previous comment. If we pass req.GetParameters() and handle everything within createAndWaitForVolume(), this parameter will have to be passed into create again. I feel it will make more sense to have this logic in createAndWaitForVolume().

internal/driver/controllerserver_helper.go
prajwalvathreya: Yup, I've removed this logic as the getAndValidateVolume() is already checking for region mismatch. This isn't needed anymore.

internal/driver/controllerserver_helper.go
komer3: Yes it will but it will just be a boolean to say either true or false. We want to move the validation check logic out from here to somewhere else where we are already doing similar checks and extracting data for making the API call. This will keep things consistent and organized

internal/driver/controllerserver_helper.go
prajwalvathreya: Look around online for better ways to do it, this seems like the most optimal method.

internal/driver/controllerserver_helper.go
```diff
@@ -403,25 +431,38 @@ func (cs *ControllerServer) validateCreateVolumeRequest(ctx context.Context, req
...
 	volumeName = preKey.GetNormalizedLabelWithPrefix(cs.driver.volumeLabelPrefix)
 	targetSizeGB = bytesToGB(size)
+	// Check if encryption should be enabled
+	if req.GetParameters()[VolumeEncryption] == True {
+		supported, err := cs.isEncryptionSupported(ctx, cs.metadata.Region)
+		if err != nil {
+			return "", targetSizeGB, size, encryptionStatus, nil
+		}
+		if !supported {
+			return "", targetSizeGB, 0, encryptionStatus, errInternal("Volume encryption is not supported in the %s region", cs.metadata.Region)
```
komer3, line 459: question: any reason for not returning volume name & size here?

internal/driver/controllerserver_helper.go
```diff
@@ -440,6 +481,10 @@ func (cs *ControllerServer) createVolumeContext(ctx context.Context, req *csi.Cr
 		volumeContext[LuksKeySizeAttribute] = req.GetParameters()[LuksKeySizeAttribute]
 	}
 
+	if req.GetParameters()[VolumeEncryption] == True {
+		volumeContext[VolumeEncryption] = True
+	}
+
```
komer3, line 487: question: are we referencing or need to use it during other steps of PVC lifecycle? Not sure if we need to set this

internal/driver/controllerserver_helper.go
```diff
@@ -448,12 +493,12 @@ func (cs *ControllerServer) createVolumeContext(ctx context.Context, req *csi.Cr
 
 // createAndWaitForVolume attempts to create a new volume and waits for it to become active.
 // It logs the process and handles any errors that occur during creation or waiting.
-func (cs *ControllerServer) createAndWaitForVolume(ctx context.Context, name string, sizeGB int, tags string, sourceInfo *linodevolumes.LinodeVolumeKey, accessibilityRequirements *csi.TopologyRequirement) (*linodego.Volume, error) {
+func (cs *ControllerServer) createAndWaitForVolume(ctx context.Context, name string, parameters map[string]string, encryptionStatus string, sizeGB int, sourceInfo *linodevolumes.LinodeVolumeKey, accessibilityRequirements *csi.TopologyRequirement) (*linodego.Volume, error) {
 	log := logger.GetLogger(ctx)
-	log.V(4).Info("Entering createAndWaitForVolume()", "name", name, "sizeGB", sizeGB, "tags", tags)
+	log.V(4).Info("Entering createAndWaitForVolume()", "name", name, "sizeGB", sizeGB, "tags", parameters[VolumeTags])
```
komer3, line 498: nit: could you add the new parameter to the log line here

docs/encrypted-drives.md
```diff
@@ -1,3 +1,77 @@
+## 🔒 Encrypted Drives
+
+**Notes**:
+
+1. **Setting Up Encryption**: In the provided StorageClasses, encryption is activated by specifying `linodebs.csi.linode.com/encrypted: "true"` in the `parameters` field. This signals the CSI driver to provision volumes with encryption enabled, provided encryption is supported in the specified region.
+2. **Retention and Expansion Options**:
+    - The `linode-block-storage-encrypted` StorageClass uses the default `Delete` reclaim policy, meaning that volumes created with this StorageClass will be deleted when the associated PVC is deleted.
+    - In contrast, the `linode-block-storage-retain-encrypted` StorageClass uses the `Retain` policy. This allows the volume to persist even after the PVC is deleted, ensuring data is preserved until manually removed.
+    - Both StorageClasses support volume expansion through the `allowVolumeExpansion: true` setting, allowing users to resize volumes as needed without data loss.
```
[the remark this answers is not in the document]
prajwalvathreya: Yup I was able to resize an encrypted volume using `linode-cli volumes resize volume-id --size number`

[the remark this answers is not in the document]
komer3: I meant try to resize using kubernetes. For example create a PVC with 10gb and then change it to 20gb. Make sure you have a pod that used the PVC as well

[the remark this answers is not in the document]
prajwalvathreya: Yup, it seems to be working when I do it in this method as well

docs/encrypted-drives.md
```diff
@@ -1,3 +1,77 @@
...
+
+3. **Default StorageClass Annotation**: By marking both StorageClasses with `storageclass.kubernetes.io/is-default-class: "true"`, they’re eligible to act as default classes. However, Kubernetes will only treat one StorageClass as the actual default. Consider applying this annotation only to the preferred default StorageClass.
+4. **Region Compatibility**: Ensure that encryption is supported in the Linode region where the volumes will be created. If encryption is not available in a specific region, the CSI driver will return an error.
```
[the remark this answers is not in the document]
prajwalvathreya: Sure

docs/encrypted-drives.md
```diff
@@ -1,3 +1,77 @@
...
+#### 🔑 Example StorageClass
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: linode-block-storage-encrypted
+  namespace: kube-system
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
```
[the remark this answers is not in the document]
prajwalvathreya: Sure

internal/driver/controllerserver_helper.go
```diff
@@ -403,25 +431,38 @@ func (cs *ControllerServer) validateCreateVolumeRequest(ctx context.Context, req
...
 
 	preKey := linodevolumes.CreateLinodeVolumeKey(0, req.GetName())
 	volumeName = preKey.GetNormalizedLabelWithPrefix(cs.driver.volumeLabelPrefix)
 	targetSizeGB = bytesToGB(size)
+	// Check if encryption should be enabled
+	if req.GetParameters()[VolumeEncryption] == True {
+		supported, err := cs.isEncryptionSupported(ctx, cs.metadata.Region)
+		if err != nil {
+			return "", targetSizeGB, size, encryptionStatus, nil
+		}
```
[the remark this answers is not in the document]
prajwalvathreya: Oh I missed returning err here. The isEncryptionSupported function returns. Updated it to err

internal/driver/controllerserver_helper.go
prajwalvathreya: No reason. I think I missed it.

internal/driver/controllerserver_helper.go
prajwalvathreya: I have set it for future purposes, just in case there is a need for it.

internal/driver/controllerserver_helper.go
```diff
@@ -403,25 +431,38 @@ func (cs *ControllerServer) validateCreateVolumeRequest(ctx context.Context, req
...
 	volumeName = preKey.GetNormalizedLabelWithPrefix(cs.driver.volumeLabelPrefix)
 	targetSizeGB = bytesToGB(size)
+	// Check if encryption should be enabled
+	if req.GetParameters()[VolumeEncryption] == True {
+		supported, err := cs.isEncryptionSupported(ctx, cs.metadata.Region)
+		if err != nil {
+			return volumeName, targetSizeGB, size, encryptionStatus, err
+		}
+		if !supported {
+			return volumeName, targetSizeGB, size, encryptionStatus, errInternal("Volume encryption is not supported in the %s region", cs.metadata.Region)
```
komer3, line 459: We have multi region support in the CSI driver so we shouldn't just use `cs.metadata.Region` here. Try to extract the region from the req. See `createLinodeVolume()` on how that is being done. 

Tbh, we should be extracting and setting the region here in this func rather than the `createLinodeVolume()`. `createLinodeVolume()` should simply take the region as parameter.

internal/driver/examples/kubernetes/csi-linode-blockstorage-encrypted.yaml
```diff
@@ -0,0 +1,44 @@
...
+kind: PersistentVolumeClaim
+metadata:
+  name: pvc-encrypted-example
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  storageClassName: linode-block-storage-encrypted
```
komer3, line 44: nit: new line here

internal/driver/controllerserver_helper.go
prajwalvathreya: I have refactored the functions for this change.

internal/driver/controllerserver_helper_test.go
```diff
@@ -477,26 +467,31 @@ func TestPrepareVolumeParams(t *testing.T) {
...
+				t.Errorf("expected error %v, got %v", tt.expectedError, err)
+			} else if err != nil && tt.expectedError != nil && err.Error() != tt.expectedError.Error() {
+				t.Errorf("expected error message %v, got %v", tt.expectedError.Error(), err.Error())
 			}
 
-			if !reflect.DeepEqual(volumeName, tt.expectedName) {
-				t.Errorf("Expected volume name: %s, but got: %s", tt.expectedName, volumeName)
-			}
+			// Only check params fields if params is not nil
+			if params != nil {
```
komer3, line 480: question: Is there a case where params will be nil?

internal/driver/controllerserver_helper_test.go
```diff
@@ -582,14 +577,24 @@ func TestPrepareVolumeParams_Encryption(t *testing.T) {
...
+			if (err != nil && tt.expectedError == nil) || (err == nil && tt.expectedError != nil) {
+				t.Errorf("expected error %v, got %v", tt.expectedError, err)
+			} else if err != nil && tt.expectedError != nil && err.Error() != tt.expectedError.Error() {
+				t.Errorf("expected error message %v, got %v", tt.expectedError.Error(), err.Error())
 			}
 
-			if encryptionStatus != tt.expectedEncrypt {
-				t.Errorf("Expected encryption status %v, got %v", tt.expectedEncrypt, encryptionStatus)
+			// Only proceed to check params fields if params is non-nil and no error was expected
+			if params != nil && err == nil {
```
komer3, line 591: same here

internal/driver/controllerserver_helper.go
```diff
@@ -489,12 +506,12 @@ func (cs *ControllerServer) createVolumeContext(ctx context.Context, req *csi.Cr
 
 // createAndWaitForVolume attempts to create a new volume and waits for it to become active.
 // It logs the process and handles any errors that occur during creation or waiting.
-func (cs *ControllerServer) createAndWaitForVolume(ctx context.Context, name string, parameters map[string]string, encryptionStatus string, sizeGB int, sourceInfo *linodevolumes.LinodeVolumeKey, accessibilityRequirements *csi.TopologyRequirement) (*linodego.Volume, error) {
+func (cs *ControllerServer) createAndWaitForVolume(ctx context.Context, name string, parameters map[string]string, encryptionStatus string, sizeGB int, sourceInfo *linodevolumes.LinodeVolumeKey, region string) (*linodego.Volume, error) {
 	log := logger.GetLogger(ctx)
 	log.V(4).Info("Entering createAndWaitForVolume()", "name", name, "sizeGB", sizeGB, "tags", parameters[VolumeTags], "encryptionStatus", encryptionStatus)
```
komer3, line 511: nit: update the log line to include new region param

internal/driver/controllerserver_helper.go
```diff
@@ -240,19 +250,10 @@ func getRegionFromTopology(requirements *csi.TopologyRequirement) string {
 
 // createLinodeVolume creates a new Linode volume with the specified label, size, and tags.
 // It returns the created volume or an error if the creation fails.
-func (cs *ControllerServer) createLinodeVolume(ctx context.Context, label, tags, encryptionStatus string, sizeGB int, accessibilityRequirements *csi.TopologyRequirement) (*linodego.Volume, error) {
+func (cs *ControllerServer) createLinodeVolume(ctx context.Context, label, tags, encryptionStatus string, sizeGB int, region string) (*linodego.Volume, error) {
 	log := logger.GetLogger(ctx)
 	log.V(4).Info("Creating Linode volume", "label", label, "sizeGB", sizeGB, "tags", tags)
```
komer3, line 255: nit: same here

internal/driver/controllerserver_helper.go
```diff
@@ -189,7 +199,7 @@ func (cs *ControllerServer) getContentSourceVolume(ctx context.Context, contentS
 // attemptCreateLinodeVolume creates a Linode volume while ensuring idempotency.
 // It checks for existing volumes with the same label and either returns the existing
 // volume or creates a new one, optionally cloning from a source volume.
-func (cs *ControllerServer) attemptCreateLinodeVolume(ctx context.Context, label, tags, volumeEncryption string, sizeGB int, sourceVolume *linodevolumes.LinodeVolumeKey, accessibilityRequirements *csi.TopologyRequirement) (*linodego.Volume, error) {
+func (cs *ControllerServer) attemptCreateLinodeVolume(ctx context.Context, label, tags, volumeEncryption string, sizeGB int, sourceVolume *linodevolumes.LinodeVolumeKey, region string) (*linodego.Volume, error) {
 	log := logger.GetLogger(ctx)
 	log.V(4).Info("Attempting to create Linode volume", "label", label, "sizeGB", sizeGB, "tags", tags)
```
komer3, line 204: nit: same here

internal/driver/controllerserver_helper.go
```diff
@@ -403,25 +432,54 @@ func (cs *ControllerServer) validateCreateVolumeRequest(ctx context.Context, req
...
 
 	preKey := linodevolumes.CreateLinodeVolumeKey(0, req.GetName())
-	volumeName = preKey.GetNormalizedLabelWithPrefix(cs.driver.volumeLabelPrefix)
-	targetSizeGB = bytesToGB(size)
+	volumeName := preKey.GetNormalizedLabelWithPrefix(cs.driver.volumeLabelPrefix)
+	targetSizeGB := bytesToGB(size)
+
+	// Check if encryption should be enabled
+	if req.GetParameters()[VolumeEncryption] == True {
+		supported, err := cs.isEncryptionSupported(ctx, cs.metadata.Region)
```
komer3, line 465: I think you forgot to use the region var you have above

internal/driver/controllerserver_helper.go
```diff
@@ -403,25 +432,54 @@ func (cs *ControllerServer) validateCreateVolumeRequest(ctx context.Context, req
...
+		if err != nil {
+			return nil, err
+		}
+		if !supported {
+			return nil, errInternal("Volume encryption is not supported in the %s region", cs.metadata.Region)
```
komer3, line 470: same here

internal/driver/controllerserver_helper_test.go
prajwalvathreya: Yup, yesterday i ran into a null pointer issue when the volume sizes were set to negative numbers. So i added this check.

internal/driver/controllerserver_helper_test.go
prajwalvathreya: same here as well

komer3 (approved):
Great work on this! Looks good to me

PR #301, opened 2024-11-06
[Bug Fix] AttachVolume error where volume region is empty

komer3:
volumeContext[VolumeTopologyRegion] was not a reliable source to get volumes region. We saw multiple instances where this was returning an empty string which was causing the region mismatch failure. Switching to just use volume obj returned by API to validate is a better and more robust approach since LinodeVolume Obj returned by the API will always have the correct region field.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/internal/driver/controllerserver.go
+++ b/internal/driver/controllerserver.go
@@ -187,7 +187,7 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 
 	// Check if the volume exists and is valid.
 	// If the volume is already attached to the specified instance, it returns its device path.
-	devicePath, err := cs.getAndValidateVolume(ctx, volumeID, instance, req.GetVolumeContext())
+	devicePath, err := cs.getAndValidateVolume(ctx, volumeID, instance)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


AshleyDumaine (approved):
LGTM

srust (approved):
What if the instance region and the volume region are actually different? How does a user recover from this situation, or get themselves into this situation?

komer3:
@srust The region mismatch error serves as a critical validation check (sort of double checking). This issue typically surfaces only in specific multi-region scenarios where cluster configuration may be incomplete.

## Primary Scenario
This issue occurs when:
- The multi-region feature is enabled. Cluster nodes lack proper region labels because of which the CSI driver cannot accurately determine the target region for volume creation

## Example Case
Consider this setup:
- Control plane with CSI driver in ORD
- Remote worker node pools in MIA and IAD
- Nodes in IAD missing region labels
- User creates a deployment with PVCs targeted for IAD

In this case, the CSI driver defaults to creating the volume in ORD (fallback region) instead of the intended IAD region, triggering the mismatch error.

**Note**: This error should only manifest when utilizing the CSI driver's multi-region capabilities.

PR #309, opened 2024-11-19
[feat] Adding Tracing(otel) to csi driver

prajwalvathreya:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


pkg/observability/tracker.go
```diff
@@ -0,0 +1,125 @@
...
+	case TracingSubfunction:
+		span.SetStatus(codes.Ok, "Sub-function call successful")
+		klog.Infof("Sub-function call in operation %s succeeded. Params: %v", operationName, params)
+	default:
+		klog.Warningf("Unknown status: %s for operation %s", status, operationName)
+	}
+}
+
+// SerializeRequest serializes an object to a JSON string for logging or processing.
+func SerializeRequest(req interface{}) string {
```
komer3, line 118: Nit: SerializeObj could be a better name here

komer3 (approved):

pkg/metrics/tracker.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	TracerProvider = trace.NewTracerProvider(
+		trace.WithBatcher(exporter),
+		trace.WithResource(res),
+	)
+
+	otel.SetTracerProvider(TracerProvider)
+
+	klog.Infof("OpenTelemetry tracing initialized for service: %s, version: %s, port: %s", serviceName, serviceVersion, tracingPort)
```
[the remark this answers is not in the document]
prajwalvathreya: Sure

.golangci.yml
```diff
@@ -119,7 +119,8 @@ linters:
     - usestdlibvars
     - varnamelen
     - whitespace
-
```
komer3, line 122: qq: reason for disabling this?

internal/driver/controllerserver_helper.go
```diff
@@ -103,6 +104,10 @@ type VolumeParams struct {
 func (cs *ControllerServer) canAttach(ctx context.Context, instance *linodego.Instance) (canAttach bool, err error) {
 	log := logger.GetLogger(ctx)
 	log.V(4).Info("Checking if volume can be attached", "instance_id", instance.ID)
+	if !observability.SkipObservability {
```
komer3, line 107: nit: could be move this skip observability check inside the StartFunctionSpan()

pkg/observability/tracker.go
```diff
@@ -0,0 +1,178 @@
...
+	tracer "go.opentelemetry.io/otel/trace"
+	"google.golang.org/grpc"
+	"k8s.io/klog/v2"
+)
+
+// Global tracing variables
+var (
+	Tracer            tracer.Tracer
+	TracerProvider    *trace.TracerProvider
+	SkipObservability = true
```
komer3, line 25: Could we just enable tracing var instead? I feel like this is just duplicated

internal/driver/controllerserver_helper.go
prajwalvathreya: I tried that, but that causes a nil pointer issue. Due to which I later discussed it with Evan and moved it here.

pkg/observability/tracker.go
prajwalvathreya: This variable is necessary for not creating spans when tracing is not required. It causes memory leaks without this variable as it's used in conditional child span creation and enabling.

.golangci.yml
prajwalvathreya: It's disabled cause it expects a defer Span.End() in the same function as StartFunctionSpan() and throws an error regarding memory leak. This sometimes causes improper span handling, which is why I explicitly end the span in the outer function calls.

docs/tracing-documentation.md
komer3, line 1: Docs looks good! But make sure also include the following:
- Setups to enable tracing for CSI driver (please make this detailed enough)
- Explain you can port-forward for dev setups
- Include the tracing on the main page README file

docs/observability.md
```diff
@@ -1,6 +1,6 @@
-# Observability with Grafana Dashboard
+# Observability for CSI Driver
 
-This document explains how to use the `grafana-dashboard` make target to install and configure observability tools, including Prometheus and Grafana, on your Kubernetes cluster. The setup uses Helm charts to install Prometheus and Grafana, provides a Prometheus data source, and applies a Grafana dashboard configuration.
+This document explains how to use the `grafana-dashboard`, `setup-tracing` make targets to install and configure observability tools.
```
komer3, line 3: nit: add an `and` between the two make targets

docs/observability.md
```diff
@@ -183,4 +183,75 @@ kubectl logs <grafana-pod-name> -n monitoring
...
+```bash
+helm template linode-csi-driver \
+	  --set apiToken="${LINODE_API_TOKEN}" \
+	  --set region="${REGION}" \
+	  --set enableTracing="true" \
+	  --set tracingPort="4318" \
+	  helm-chart/csi-driver --namespace kube-system > csi.yaml
+```
+
+### 2. Delete the Existing Release of the CSI Driver
```
komer3, line 205: nit: Include if they have it installed. Some people might not

hack/setup-tracing.sh
```diff
@@ -0,0 +1,52 @@
...
+
+echo "Waiting for Jaeger LoadBalancer to get an external IP..."
+EXTERNAL_IP=""
+while [[ -z "$EXTERNAL_IP" ]]; do
+  EXTERNAL_IP=$(get_jaeger_lb_ip)
+  if [[ -z "$EXTERNAL_IP" ]]; then
+    echo "Waiting for LoadBalancer external IP..."
+    sleep 10
+  fi
+done
```
komer3, line 46: question: Is this going infinite loop? May be we do set amount of retries or have some backoff retry logic here?

prajwalvathreya: I've updated the loop with number of retires before backing off

komer3 (approved):
Great work on this! LGTM

PR #312, opened 2024-11-21
refactor: remove capacity checking when attaching

guilhem:
This pull request focuses on simplifying the volume attachment logic in the `ControllerServer` by removing redundant checks and related methods. The most important changes include the removal of the `checkAttachmentCapacity` function and its associated methods, as well as the related test cases.

### Context

[CSI caller (like Kubernetes) should already respect](https://kubernetes-csi.github.io/docs/volume-limits.html) volume limitation from `NodeGetInfo`.
We don't need another check and logic at attach.

### Motivation

Reduce logic and complexity and rely more on Linode API calls.
If there is an error, return it to the user for easier debug.

### Risk

Making more failing attach requests and maybe hitting a rate limit.
This should not be a problem, [as CSI attacher have a backoff logic](https://github.com/kubernetes-csi/external-attacher?tab=readme-ov-file#csi-error-and-timeout-handling).
Moreover, current limit logic is already doing api calls that should have hit API ratelimit.

### Possible Mitigations

* Implementing a rate limiting on api call.
* [As GCP,](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/a28f8d39b21ca8439d64445da33ed3d54a4dc67c/pkg/gce-pd-csi-driver/controller.go#L158C6-L158C21) implementing a `csiErrorBackoff` mechanism

IMHO, both are not mandatory for the moment.

### Simplification of volume attachment logic:

* Removed the `checkAttachmentCapacity` function and its invocation from the `ControllerPublishVolume` method in `internal/driver/controllerserver.go`.
* Deleted the `canAttach` function, which was used to determine if an additional volume could be attached to an instance, from `internal/driver/controllerserver_helper.go`.
* Removed the `checkAttachmentCapacity` function, which checked if an instance could accommodate additional volume attachments, from `internal/driver/controllerserver_helper.go`.

### Removal of related test cases:

* Deleted the `TestCheckAttachmentCapacity` test function from `internal/driver/controllerserver_helper_test.go`, which tested the `checkAttachmentCapacity` logic.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


internal/driver/controllerserver_helper.go
```diff
@@ -95,31 +95,6 @@ type VolumeParams struct {
...
-		return false, errInternal("list instance volumes: %v", err)
-	}
-
-	// Return true if the number of attached volumes is less than the limit
-	return len(volumes) < limit, nil
-}
-
 // maxAllowedVolumeAttachments calculates the maximum number of volumes that can be attached to a Linode instance,
 // taking into account the instance's memory and currently attached disks.
 func (cs *ControllerServer) maxAllowedVolumeAttachments(ctx context.Context, instance *linodego.Instance) (int, error) {
```
komer3, line 100: You can also remove this helper func(). Its not used anywhere else except the funcs you are removing :)

komer3 (commented):
I agree with this proposal. This will definitely simplify the volume attachment logic (great for us when debugging and understanding what is happening).   

Just to make sure we aren't missing any cases, we should test the CSI driver by running the all the e2e tests we have plus run manual test to see how CSI driver handles things when we hit the max limit.

wbh1:
@guilhem - do you intend to re-open https://github.com/linode/linode-blockstorage-csi-driver/pull/310? I believe those changes will still be beneficial in order to properly implement the CSI spec and behave the way that kube-scheduler expects.

guilhem:
duplicate #322

PR #322, opened 2024-11-25
Simpleattach

guilhem:
This pull request focuses on simplifying the volume attachment logic in the `ControllerServer` by removing redundant checks and related methods. The most important changes include the removal of the `checkAttachmentCapacity` function and its associated methods, as well as the related test cases.

### Context

[CSI caller (like Kubernetes) should already respect](https://kubernetes-csi.github.io/docs/volume-limits.html) volume limitation from `NodeGetInfo`.
We don't need another check and logic at attach.

### Motivation

Reduce logic and complexity and rely more on Linode API calls.
If there is an error, return it to the user for easier debug.

### Risk

Making more failing attach requests and maybe hitting a rate limit.
This should not be a problem, [as CSI attacher have a backoff logic](https://github.com/kubernetes-csi/external-attacher?tab=readme-ov-file#csi-error-and-timeout-handling).
Moreover, current limit logic is already doing api calls that should have hit API ratelimit.

### Possible Mitigations

* Implementing a rate limiting on api call.
* [As GCP,](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/a28f8d39b21ca8439d64445da33ed3d54a4dc67c/pkg/gce-pd-csi-driver/controller.go#L158C6-L158C21) implementing a `csiErrorBackoff` mechanism

IMHO, both are not mandatory for the moment.

### Simplification of volume attachment logic:

* Removed the `checkAttachmentCapacity` function and its invocation from the `ControllerPublishVolume` method in `internal/driver/controllerserver.go`.
* Deleted the `canAttach` function, which was used to determine if an additional volume could be attached to an instance, from `internal/driver/controllerserver_helper.go`.
* Removed the `checkAttachmentCapacity` function, which checked if an instance could accommodate additional volume attachments, from `internal/driver/controllerserver_helper.go`.

### Removal of related test cases:

* Deleted the `TestCheckAttachmentCapacity` test function from `internal/driver/controllerserver_helper_test.go`, which tested the `checkAttachmentCapacity` logic.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


internal/driver/controllerserver_helper.go
```diff
@@ -719,13 +665,17 @@ func (cs *ControllerServer) attachVolume(ctx context.Context, volumeID, linodeID
...
-		// Check if the error indicates that the volume is already attached.
+		// https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerpublishvolume-errors
 		var apiErr *linodego.Error
-		if errors.As(err, &apiErr) && strings.Contains(apiErr.Message, "is already attached") {
-			code = codes.Unavailable // Allow a retry if the volume is already attached: race condition can occur here
+		if errors.As(err, &apiErr) {
+			switch {
+			case strings.Contains(apiErr.Message, "is already attached"):
+				return errAlreadyAttached
+			case strings.Contains(apiErr.Message, "Maximum number of block storage volumes are attached to this Linode"):
```
srust, line 674: Ok, but this seems brittle when the API changes their text message. How can we check this error, but not rely on the spelling of an API text error message?

internal/driver/controllerserver_helper.go
```diff
@@ -677,33 +650,6 @@ func (cs *ControllerServer) getInstance(ctx context.Context, linodeID int) (*lin
 	return instance, nil
 }
```
srust, line 652: Is there really no value in having these checks client side at all?

internal/driver/controllerserver_helper.go
guilhem: That's an excellent question.
API doesn't document errors
image
And linodego [report request as it](https://github.com/linode/linodego/blob/bb8fd62f542430bc465a75f3ba9a00293680db4b/volumes.go#L127C1-L131C2)

But we have to notice this logic was already done before in CSI :/
https://github.com/linode/linode-blockstorage-csi-driver/blob/17b7d1b66edf6762fabbc98c0ad944c417c309a7/internal/driver/controllerserver_helper.go#L725

internal/driver/controllerserver_helper.go
guilhem: I tend to prefer to rely on API internal logic and validation and for CSI to act as a simple pass-through.

The only problem could be a rate-limiting issue on API (GET vs. POST).
But attacher already has a backoff on error
https://github.com/kubernetes-csi/external-attacher?tab=readme-ov-file#csi-error-and-timeout-handling

But that may be questionable, or we may have to wait for linodego to implement rate-limiting.
That's why this PR was in draft state at first.

internal/driver/controllerserver_helper.go
```diff
@@ -719,13 +665,17 @@ func (cs *ControllerServer) attachVolume(ctx context.Context, volumeID, linodeID
...
-		code := codes.Internal // Default error code is Internal.
-		// Check if the error indicates that the volume is already attached.
+		// https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerpublishvolume-errors
 		var apiErr *linodego.Error
-		if errors.As(err, &apiErr) && strings.Contains(apiErr.Message, "is already attached") {
-			code = codes.Unavailable // Allow a retry if the volume is already attached: race condition can occur here
+		if errors.As(err, &apiErr) {
+			switch {
+			case strings.Contains(apiErr.Message, "is already attached"):
+				return errAlreadyAttached
```
komer3, line 673: Question: We only get this err when the volume is attached to another linode? 

I think pass the specific err message from the api which could include linode id of another node. This would be helpful in debugging issue in future.

internal/driver/controllerserver_helper.go
```diff
@@ -719,13 +665,17 @@ func (cs *ControllerServer) attachVolume(ctx context.Context, volumeID, linodeID
...
+			case strings.Contains(apiErr.Message, "Maximum number of block storage volumes are attached to this Linode"):
+				return errMaxAttachments
```
komer3, line 675: Similar to my suggestion from other comment. We should try to pass along the err message by the api here too

internal/driver/controllerserver_helper.go
komer3: Yeah def not the ideal way to do it. Hopefully linodego can provide better error codes in future so we don't have to build logic based on the err strings :/

PR #328, opened 2024-11-26
feat: implement GET_VOLUME

guilhem:
Report more information in volumeCondition for GET and LIST.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/internal/driver/controllerserver.go
+++ b/internal/driver/controllerserver.go
@@ -377,37 +377,12 @@ func (cs *ControllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
 
 	entries := make([]*csi.ListVolumesResponse_Entry, 0, len(volumes))
 	for volNum := range volumes {
-		key := linodevolumes.CreateLinodeVolumeKey(volumes[volNum].ID, volumes[volNum].Label)
-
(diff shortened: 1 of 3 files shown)


internal/driver/controllerserver.go
```diff
@@ -493,3 +468,82 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
...
+	if volume.LinodeID != nil {
+		publishedNodeIds = append(publishedNodeIds, strconv.Itoa(*volume.LinodeID))
+	}
+
+	csiVolume = &csi.Volume{
+		VolumeId:      key.GetVolumeKey(),
+		CapacityBytes: gbToBytes(volume.Size),
+	}
+
+	if volume.Region != "" {
```
komer3, line 528: Do we need this check here? Do you think we will have a case where linode api will return a volume obj with empty region?

guilhem: It should not happen with the current volume implementation.
But I tend to validate any input 😅 
But I tried to keep it as logical as possible and not pushing empty string region.

komer3: Thats fair

komer3 (approved):

PR #360, opened 2025-01-27
Replacing init container (In progress)

komer3:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


helm-chart/csi-driver/templates/daemonset.yaml
```diff
@@ -121,10 +94,6 @@ spec:
       volumes:
       - emptyDir: {}
         name: linode-info
```
AshleyDumaine, line 96: we don't need this anymore right?

komer3: yes. good catch!

srust:
Hi @komer3 , the `get-linode-id.sh` script uses metadata, then dmidecode, then falls back to the API / kubernetes as a last step. Because of this, it requires CAP_SYS_ADMIN and also host networking.

The changes proposed remove these first priority steps. Are we sure that the capabilities and appropriate fallbacks are equivalent with this change?

internal/driver/metadata.go
```diff
@@ -55,19 +83,53 @@ func GetNodeMetadata(ctx context.Context, cloudProvider linodeclient.LinodeClien
...
+	if err != nil {
+		return Metadata{}, fmt.Errorf("failed to get node %s: %w", nodeName, err)
+	}
+
+	providerID := node.Spec.ProviderID
+	if providerID == "" {
+		return Metadata{}, fmt.Errorf("provider ID not found for node %s", nodeName)
+	}
+
+	// Extract Linode ID from provider ID (format: linode://12345)
```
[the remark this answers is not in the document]
komer3: Reason for removing that was on the assumption that we will all always run CSI driver and CCM together. CCM will add the provider ID to the node resource. This way we are limiting exec calls to the OS. 

Is my assumption correct? Or do we have scenarios where we run CSI driver without running CCM?

komer3:
> Hi @komer3 , the `get-linode-id.sh` script uses metadata, then dmidecode, then falls back to the API / kubernetes as a last step. Because of this, it requires CAP_SYS_ADMIN and also host networking.
> 
> The changes proposed remove these first priority steps. Are we sure that the capabilities and appropriate fallbacks are equivalent with this change?

Hi @srust, With this change we are removing demicode from the steps on the assumption that when we run CSI driver we also run CCM on the cluster (which would populate the provider id on the node resource). If this is not the case and we have scenarios where we would want to run CSI driver without running CCM, then we would definitely want demicode step. I wanted to reduce step if they aren't needed. 

I'm open to adding demicode back as another safety fall back step.

helm-chart/csi-driver/templates/daemonset.yaml
komer3: Removed!

internal/driver/metadata.go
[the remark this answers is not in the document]
tchinmai7: we've seen some cases where rke2 clusters had `rke2://` as the provider-id - i'd be cool if the csi driver still continues to support those cases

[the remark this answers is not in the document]
komer3: Oh nice. That good to know!

Not sure if we need to keep demicode still. The provider ID is only needed for node server which has host network access and gets the provider ID from metadata service. Controller server doesn't need to get the provider ID (it doesn't use it for anything). 

Has you seen issues with accessing metadata service on RKE2 nodes before?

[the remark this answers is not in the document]
komer3: @tchinmai7 I add the step where we check for the product serial (which equals to Linode ID) back. Let me know if you have any other suggestions

.github/workflows/ci.yml
```diff
@@ -35,6 +35,8 @@ jobs:
     steps:
       - name: Install cryptsetup
         run: |
+          sudo apt-get update
+          sudo apt-get install -f
```
tchinmai7, line 39: do we need `install -f `?

tchinmai7 (dismissed):

.github/workflows/ci.yml
komer3: I think we do honestly. Just put it there to be safe. CI kept failing randomly. I can remove it and try

tchinmai7 (approved):

PR #362, opened 2025-02-03
[Fix] LUKS device on reattach not executing luksOpen

komer3:
### Reason for the change:

We had report on an issue when we try to remount the luks volume. We would get stuck at the validation part (volumeName) when it is expected to not have volume name on reattach. This change ignores that specific check if we are remounting a LUKS volume.

For more info and how to recreate the bug: https://github.com/ErjanGavalji/linode-luks-pv-recreation-error

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/internal/driver/nodeserver_helpers.go
+++ b/internal/driver/nodeserver_helpers.go
@@ -16,6 +16,7 @@ limitations under the License.
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


internal/driver/nodeserver_luks_encryption_test.go
```diff
@@ -33,7 +33,7 @@ func TestNodeServer_mountVolume_luks(t *testing.T) {
 			name:       "Success - mount LUKS volume",
 			devicePath: "/tmp/test",
 			req: &csi.NodeStageVolumeRequest{
-				VolumeId: "test",
+				VolumeId: "123-test",
```
AshleyDumaine, line 36: nit: we might want to make this a variable so we don't need to set it in each test

komer3: Yeah I can do that!

internal/driver/nodeserver_helpers.go
```diff
@@ -280,6 +280,12 @@ func (ns *NodeServer) mountVolume(ctx context.Context, devicePath string, req *c
 
 	stagingTargetPath := req.GetStagingTargetPath()
 	volumeCapability := req.GetVolumeCapability()
+	volumeID := req.GetVolumeId()
```
AshleyDumaine, line 283: should we add a check here to make sure it's not an empty string?
```
	if volumeID == "" {
		return errNoVolumeID
	}
```

komer3: I can add a test case for that

komer3: We don't necessary need to add that since the getMountSource() will throw an error if its empty anyway

AshleyDumaine (approved):

PR #363, opened 2025-02-04
set securityContext for Pods and containers

AshleyDumaine:
This explicitly sets the securityContext on the Pods and containers and tries to restrict permissions where possible for privileged containers. Specifically, the csi-linode-plugin container needed to be allowed due to the Bidirectional volume mount propagation requirement but the other containers were set to not allow privileged permissions.
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/helm-chart/csi-driver/templates/daemonset.yaml
+++ b/helm-chart/csi-driver/templates/daemonset.yaml
@@ -15,6 +15,9 @@ spec:
         app: csi-linode-node
         role: csi-linode
     spec:
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
       containers:
       - args:
         - --v=2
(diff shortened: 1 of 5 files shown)


internal/driver/deploy/releases/linode-blockstorage-csi-driver.yaml
```diff
@@ -357,10 +380,6 @@ spec:
       volumes:
       - emptyDir: {}
         name: socket-dir
-      - configMap:
```
AshleyDumaine, line 360: I'm not sure why this was still here but I would think it's good to remove since https://github.com/linode/linode-blockstorage-csi-driver/pull/360 was merged?

komer3: oh yeah. Thanks for catching that. We want to remove it

komer3 (dismissed):

helm-chart/csi-driver/templates/daemonset.yaml
komer3, line 1: question: we need to add the security contexts for statefulset manifest as well right?

https://github.com/linode/linode-blockstorage-csi-driver/blob/main/helm-chart/csi-driver/templates/csi-linode-controller.yaml

AshleyDumaine: yep, good catch

komer3 (approved):

PR #382, opened 2025-03-07
Resize fs

guilhem:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/internal/driver/identityserver.go
+++ b/internal/driver/identityserver.go
@@ -105,7 +105,7 @@ func (linodeIdentity *IdentityServer) GetPluginCapabilities(ctx context.Context,
 					// 	2. Delete and recreate the pod that is using the PVC(or scale replicas accordingly)
 					// 	3. This operation should detach and re-attach the volume to the newly created pod allowing you to use the updated size
 					VolumeExpansion: &csi.PluginCapability_VolumeExpansion{
-						Type: csi.PluginCapability_VolumeExpansion_ONLINE,
+						Type: csi.PluginCapability_VolumeExpansion_OFFLINE,
 					},
 				},
 			},
--- a/internal/driver/nodeserver.go
+++ b/internal/driver/nodeserver.go
@@ -365,7 +365,7 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	log.V(4).Info("Processed LinodeVolumeKey", "LinodeVolumeKey", LinodeVolumeKey)
 	if err != nil {
 		observability.RecordMetrics(observability.NodeExpandTotal, observability.NodeExpandDuration, observability.Failed, functionStartTime)
-		return nil, errors.Join(status.Errorf(codes.InvalidArgument, "volume not found: %v", err), err)
+		return nil, errors.Join(status.Errorf(codes.NotFound, "volume not found: %v", err), err)
 	}
 
 	// We have no context for the partition, so we'll leave it empty
--- a/internal/driver/nodeserver.go
+++ b/internal/driver/nodeserver.go
@@ -15,7 +15,6 @@ limitations under the License.
 */
 
 import (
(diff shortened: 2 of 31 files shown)


internal/driver/nodeserver_helpers.go
```diff
@@ -35,6 +36,11 @@ const (
 	ownerGroupReadWritePermissions = os.FileMode(0o660)
 )
 
+var (
+	ErrNoVolumeCapability = errors.New("volume capability is required")
+	ErrNoAccessMode       = errors.New("access mode is nil")
+)
+
```
komer3, line 43: nit: could you move these to errors.go

internal/driver/nodeserver.go
```diff
@@ -411,3 +478,18 @@ func (ns *NodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
...
+	}
+
+	if needResize {
+		if _, err := ns.resizeFs.Resize(devicePath, volumePath); err != nil {
+			return false, fmt.Errorf("could not resize volume: %w", err)
+		}
+	}
+
+	return needResize, nil
+}
```
komer3, line 495: nit: could you move this to the helper file?

internal/driver/nodeserver.go
```diff
@@ -256,14 +275,27 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+		resized, err := ns.resize(devicePath, stagingTargetPath)
+		if err != nil {
+			observability.RecordMetrics(observability.NodeStageVolumeTotal, observability.NodeStageVolumeDuration, observability.Failed, functionStartTime)
+			return nil, errInternal("failed to resize volume %s: %v", volumeID, err)
+		}
+		if resized {
+			log.V(4).Info("Successfully resized volume", "volumeID", volumeID)
+		}
+	}
+
```
komer3, line 298: Do we still need this here? Resize operation should be executed in `NodeExpandVolume()` right after this func is executed

internal/driver/identityserver.go
```diff
@@ -105,7 +105,7 @@ func (linodeIdentity *IdentityServer) GetPluginCapabilities(ctx context.Context,
 					// 	2. Delete and recreate the pod that is using the PVC(or scale replicas accordingly)
 					// 	3. This operation should detach and re-attach the volume to the newly created pod allowing you to use the updated size
 					VolumeExpansion: &csi.PluginCapability_VolumeExpansion{
-						Type: csi.PluginCapability_VolumeExpansion_ONLINE,
+						Type: csi.PluginCapability_VolumeExpansion_OFFLINE,
```
komer3, line 108: nit: could update the docs to reflect how resize operations work in CSI?

internal/driver/nodeserver.go
guilhem: It should be in both.
Customers could extend themselves the PV without noticing Kubernetes :)

As it's a really idempotent action (even checked if "needResize"), this is not really a problem :)

In future, if we have online resizing, this part will be only for "offline".

internal/driver/identityserver.go
guilhem: doc just above was good :)
I added a line in /docs

internal/driver/nodeserver.go
komer3: Ah so its for the case where customer changes the volume size manually from cloud manager. But do we need to consider that case? Shouldn't customer just do it by updating PVC?

guilhem: In fact, all big CSI I was looking at was doing the same. 
So in fact I just tried to keep the same experience. 

In a previous life, I did use a lot of terraform to provision and change PV.
Expecting Kubernetes to deal with it :)

komer3 (approved):
LGTM!

PR #414, opened 2025-04-21
[WIP] Remove linode api token requirement from node-server

jescarri:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/internal/driver/nodeserver.go
+++ b/internal/driver/nodeserver.go
@@ -500,13 +500,13 @@ func (ns *NodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
 	// devices that can be attached.
 	log.V(4).Info("Listing attached block devices", "nodeID", ns.metadata.ID)
 
-	lsblkOutput, err := execRunner("lsblk", "-J", "--nodeps")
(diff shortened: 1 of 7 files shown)


internal/driver/nodeserver.go
```diff
@@ -469,6 +479,10 @@ func (ns *NodeServer) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetC
 	}, nil
 }
 
+func execRunner(name string, arg ...string) ([]byte, error) {
+	return exec.Command(name, arg...).CombinedOutput()
+}
+
```
komer3, line 485: Could we use the interfaces defined for exec calls here: https://github.com/linode/linode-blockstorage-csi-driver/blob/d43f82da3e16f6294f1b1cf32eaa3d621c53d044/pkg/mount-manager/safe_mounter.go

This would be helpful during mocking for unit testing

Could you also move the structs and  execRunner() to either helper file _OR_ somewhere in the pkg directory?

internal/driver/nodeserver.go
komer3, line 1: What are your thoughts on using this library instead of exec command? 

https://github.com/jaypipes/ghw?tab=readme-ov-file#block-storage

PR #426, opened 2025-05-16
fix: persist pv on reboot

guilhem:
CSI don't detach pv on node reboot.
csiattachement still exists even if pv have been removed. Persist on reboot permit to have same behavior as all other clouds.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


eljohnson92:
I am going to close this PR as setting this flag to persist would be a regression in the number of volumes that are supported on most nodes(it would limit us to only 7 volumes per node in all cases). we can open up a new PR for this issue in the future if things change

PR #570, opened 2026-04-08
fix(deps): bump gRPC to v1.80.0 and CSI sidecars to latest

komer3:
## Summary

- Upgrade `google.golang.org/grpc` from v1.79.3 to v1.80.0
- Bump all CSI sidecar container images to their latest versions:
  - **csi-provisioner**: v5.3.0 → v6.2.0
  - **csi-attacher**: v4.9.0 → v4.11.0
  - **csi-resizer**: v1.14.0 → v2.1.0
  - **csi-node-driver-registrar**: v2.14.0 → v2.16.0
- Updates applied to both kustomize base manifests and Helm chart values

> **Note:** None of the latest sidecar releases include gRPC >= v1.79.3 (the version that fixes the reported CVE). The closest is csi-provisioner v6.2.0 which uses gRPC v1.79.1. The other sidecars use v1.72.2–v1.78.0. The upstream CSI sidecar projects have not yet bumped their gRPC dependency to include the fix.

## Test plan

- [ ] Verify CSI driver deploys successfully with updated sidecar versions
- [ ] Test volume provisioning, attaching, resizing, and node registration
- [ ] Confirm no regressions from major version bumps (csi-provisioner v5→v6, csi-resizer v1→v2)


moshevayner:
Hey @komer3 👋 
Looks like we have 2 major version bumps there (provisioner and resizer). Do we know what are the breaking changes and if we might be impacted by them?

komer3:
We have CSI-Sanity test suite that runs a rigorous test suite e2e. If they all pass, we should not be impacted!

moshevayner (approved):

amold1 (approved):

PR #573, opened 2026-04-30
[ci] enhance workflows with hardened runner and updated actions

moshevayner:
Signed-off-by: Moshe Vayner <moshe@vayner.me>

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -5,21 +5,22 @@ on:
   pull_request_target:
     types: [opened, reopened, synchronize]
 
+permissions:
+  contents: read
+
 jobs:
   label-pr:
     name: Update PR labels
     permissions:
-      contents: write
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v6
+      - name: Harden Runner
+        uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
         with:
-          fetch-depth: 0
+          disable-sudo: true
+          egress-policy: audit
       - name: Label PR
-        uses: release-drafter/release-drafter@v7
-        with:
-          disable-releaser: github.ref != 'refs/heads/main'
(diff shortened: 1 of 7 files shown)


.github/workflows/pr-labeler.yml
```diff
@@ -5,21 +5,22 @@ on:
...
+      - name: Harden Runner
+        uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
         with:
-          fetch-depth: 0
+          disable-sudo: true
+          egress-policy: audit
       - name: Label PR
-        uses: release-drafter/release-drafter@v7
-        with:
-          disable-releaser: github.ref != 'refs/heads/main'
```
komer3, line 23: Is it safe to remove disable-releaser flag?

.github/workflows/pr-labeler.yml
```diff
@@ -5,21 +5,22 @@ on:
...
     types: [opened, reopened, synchronize]
 
+permissions:
+  contents: read
+
 jobs:
   label-pr:
     name: Update PR labels
     permissions:
-      contents: write
```
komer3, line 12: Save to remove the contents: write ?

.github/workflows/pr-labeler.yml
moshevayner: With the new version- it's no longer supported. They removed this flag:
https://github.com/release-drafter/release-drafter/releases/tag/v7.1.1
So with the bump to the latest version, my linter complained that this flag is no longer supported.

.github/workflows/pr-labeler.yml
moshevayner: According to their docs, the `autolabeler` job only needs `contents: read` and `pull-requests: write`, so I followed their guide here:
https://github.com/release-drafter/release-drafter#autolabeler

I think `contents: write` was a bit over-privileged for what it needs to do, because it only adds labels to PRs. Does that make sense?

komer3: Yup sounds good. Just wanted to make sure so I can do the same for CAPL workflow!

komer3 (approved):
LGTM! We can do a CSI driver release after this is merged to audit the API calls for release and helm workflows

PR #578, opened 2026-05-14
[fix] Prefer PublishContext device paths in NodeStageVolume

komer3:
## Summary

This change updates `NodeStageVolume` to prefer the controller-provided `PublishContext.devicePath` when it exists on the current node, and to fall back to the existing by-id lookup when it does not.

It also adds targeted `NodeStageVolume` coverage for both the PublishContext-first path and the fallback path, using Linode-style  device paths.

## Why

Older volumes can carry a persisted CSI `volumeHandle` that reflects a historical label format, while the actual Linode volume label may differ. In that situation, node-side by-id resolution derived from the volume handle can fail even though the controller has already published the correct attached device path for the current node.

By checking `PublishContext.devicePath` first, `NodeStageVolume` can stage volumes using the device path returned by the controller during attach. If that device path is not present on the node, the code still falls back to the previous by-id lookup behavior.

That makes the staging path more resilient for older volumes with inconsistent label versus volume-handle history, without removing the existing fallback.

## Validation

- ran `make test`

### DIFF
--- a/internal/driver/nodeserver.go
+++ b/internal/driver/nodeserver.go
@@ -237,8 +237,8 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		}
 	}
 
-	log.V(4).Info("Finding device path", "volumeID", volumeID)
-	devicePath, err := ns.findDevicePath(ctx, *LinodeVolumeKey, partition)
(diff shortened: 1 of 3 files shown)


moshevayner (approved):
LGTM! 🚀

PR #596, opened 2026-07-23
[CI] group renovate security PRs

AshleyDumaine:
Adds the missing groupName like we have in CAPL: https://github.com/linode/cluster-api-provider-linode/blob/main/renovate.json5#L15

This will result in opening all vulnerability updates in one PR like so: https://github.com/linode/cluster-api-provider-linode/pull/1104

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,7 @@ jobs:
         with:
           base: ${{ github.ref }}
           filters: .github/filters.yml
+          predicate-quantifier: 'every'
   ci:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -34,6 +34,7 @@ jobs:
         with:
           base: ${{ github.ref }}
           filters: .github/filters.yml
+          predicate-quantifier: 'every'
```
moshevayner, line 37: TIL!

moshevayner (approved):

